PR #72, opened 2026-08-02
fix(installer): SDDM login theme silently fails to install in the default TUI

XephyLon:
## Problem

Selecting **SDDM login theme** in the installer silently installed nothing.

`ii-sddm-theme`'s upstream `setup.sh` is purely interactive — its first action is a
`read "Continue? [y/n]"`, and it has no non-interactive / env-driven mode. The default
fzf TUI's **quiet mode** runs the whole pipeline as `setup install --force </dev/null`,
so `5.sddm-theme.sh` handed off to the upstream installer with stdin on `/dev/null`. That
first `read` hit EOF, fell through to its default case, and the installer `exit 0`'d —
installing nothing, with no error.

Wallpaper Engine and Plymouth survive `</dev/null` because they never prompt; SDDM was the
only interactive step. The whiptail fallback and the fzf *verbose* path were unaffected
(real terminal), which is why the failure looked inconsistent.

## Fix

**`tui.sh`** — stop exporting `INSTALL_SDDM` into the quiet pipeline; run `5.sddm-theme.sh`
as a post-install step on the clean terminal after the progress bar finishes, mirroring how
the existing fcitx5 extra is already handled.

**`5.sddm-theme.sh`** — defense-in-depth: when stdin isn't a TTY, reconnect the upstream
installer to `/dev/tty` so its prompts still reach the user; if there's no controlling
terminal at all (headless/CI), skip loudly with guidance instead of aborting silently. This
hardens every other caller of the wrapper too.

## Testing

- `bash -n` clean on both scripts.
- Verified the three-way stdin guard selects the correct branch (interactive → run directly;
  non-TTY stdin with a controlling terminal → reconnect to `/dev/tty`; no terminal at all →
  skip loudly with `exit 0`, no silent abort).

### DIFF
--- a/sdata/subcmd-install/5.sddm-theme.sh
+++ b/sdata/subcmd-install/5.sddm-theme.sh
@@ -48,6 +48,23 @@ if ! curl -fsSL "$SETUP_URL" -o "$TMP_SETUP"; then
 fi
 
 echo "[ImI] SDDM theme: handing off to the upstream installer (interactive)..."
+# The upstream installer is purely interactive: its very first action is a
+# `read -p "Continue? [y/n]"`, and it has no non-interactive/env-driven mode.
+# If our stdin is not a terminal, that `read` gets EOF, falls to its default
+# case, and the installer exits 0 — silently installing nothing. This bites the
+# quiet install path, which runs the whole pipeline with stdin on /dev/null.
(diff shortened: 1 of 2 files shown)


XephyLon:

PR #74, opened 2026-08-02
NixOS flake: root flake, home-manager module, and a VM-tested three-tier split

XephyLon:
Implements the first pass of the NixOS flake proposal this PR has tracked since the draft. The proposal doc rides in the branch and now describes what exists rather than what might.

What's in:

- **Root `flake.nix`** — `packages.<system>.immaterial-impulse` (the shell tree, byte-verbatim, as a store path), `quickshell-wrapped` (shared with the nested dist-nix flake via a new `qsPackage` argument instead of being duplicated), and `devShells.default`.
- **`nix/hm-module.nix`** — the home-manager module implementing the settled three-tier split: the QML tree and matugen config + templates symlinked from the store; `config.json`, `plugin-state.json` and `plugins/` seeded by activation only when absent, the user's from first write onward; matugen **outputs** never managed, with a warning when the surrounding home-manager configuration manages one.
- **`sdata/dist-nix` reconciled** — `--via-nix` marked superseded in its README; the nested flake stays (it is the whole working non-NixOS standalone home-manager path), and `quickshell.nix` is imported by the root flake in place.
- **`flake.lock` committed** — nixpkgs at nixos-25.11; quickshell pinned in the lock to the PKGBUILD's `_commit` (e649d247, the commit the shell is built and tested against everywhere else), with `flake.nix` naming the PKGBUILD as the source of truth for re-pins. The hardcoded-commit pin is gone from the URL.
- `nixosModules.default` is **omitted, not stubbed** — home-manager alone covers the shell; the system half (SDDM, plymouth, polkit) stays under open questions.

Validated (details in the doc's "Validated so far"):

- `nix flake check` green. The first run caught a real bug — fixupPhase's shebang patcher aborts on `generate_colors_material.py`'s `env -S` venv shebang — fixed by shipping the tree verbatim (`dontPatchShebangs`).
- `nix build .#immaterial-impulse` green; output tree verified byte-identical.
- A consumer-side home-manager eval builds a full generation, and a probe config managing `gtk-3.0/gtk.css` makes the module emit exactly the tier-3 warning.
- A qemu/KVM NixOS VM test: a guest with the module enabled boots, both tier-1 symlinks resolve into a read-only store, `config.json` seeds byte-equal to the defaults, a user overwrite of it survives a home-manager service restart, no matugen output is store-managed, and `qs --version` runs from the user profile.

Review round (all three must-fix items addressed):

- quickshell re-pinned to the PKGBUILD's commit (above); `nix flake check` and `nix build` re-run green on the new lock.
- The venv hole is **named, not fixed**: the doc's "Validated so far" now states that the module ships a shell that cannot generate colours — eight scripts need the uv venv and `IMMATERIAL_IMPULSE_VIRTUAL_ENV`, and closing it is its own follow-up because `requirements.txt` pins three packages nixpkgs' python set does not carry.
- README gains a "NixOS (experimental)" install subsection naming the flake input, the module, and what does not work yet.
- Nits: branch name in the doc header corrected, the `4b43790` citation carries its subject, `.gitignore` ignores `result`, and the package filters `__pycache__`/`.pytest_cache` out of the source (proven against a real cache in the tree).

Still open, named in the doc: the venv for colour generation (first blocker of a usable shell), running the Wayland session from the store path, `run_tests.sh` under `nix develop`, runtime plugin install under an immutable tree, the `exp-update` replacement, the `scripts/` PATH enumeration, CI `nix flake check`, and the matugen-output-list drift guard.

Changelog: not user-visible — nothing under dots/ changes in this PR
Docs: updated docs/proposals/nixos-flake.md §"Sketch", §"Validated so far", §"Open questions"; README.md §"Install / update"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ __pycache__/
 /os-release
 # Emacs auto backup file
 *~
+# nix build leaves a result symlink at the root
+/result
--- a/README.md
+++ b/README.md
@@ -16,6 +16,26 @@ The installer is idempotent - re-running it updates an existing install
 components: Wallpaper Engine integration, the imi-sddm-theme login theme,
 and a Plymouth boot splash.
 
+### NixOS (experimental)
+
+The repository root is a Nix flake. Instead of the installer, a
+home-manager user adds:
+
+```nix
+inputs.immaterial-impulse.url = "github:XephyLon/immaterial-impulse";
+```
+
+and imports `inputs.immaterial-impulse.homeManagerModules.default` with
+`programs.immaterial-impulse.enable = true`. The module symlinks the
+shell tree and the matugen config from the store, seeds
+`~/.config/immaterial-impulse` once, and refuses to manage matugen's
+generated outputs (it warns if your home-manager config does).
+
+What does not work yet: wallpaper-driven colour generation (the scripts
+need a Python venv the module does not provide) and the Wallpaper Engine
+integration. Design, validation status, and open questions are in
+[docs/proposals/nixos-flake.md](docs/proposals/nixos-flake.md).
+
 ## Compositor support
 
 **Hyprland only.** There are no plans to support Niri or any other
--- a/docs/proposals/nixos-flake.md
+++ b/docs/proposals/nixos-flake.md
@@ -1,6 +1,11 @@
 # Proposal: NixOS flake
 
-> Draft / tracking proposal. Not scheduled.
+> In progress. The first pass is implemented on `feat/nixos-flake`: a root
+> `flake.nix` with `packages.<system>.immaterial-impulse`,
+> `homeManagerModules.default` and `devShells.default`, the three-tier
+> declarative/mutable split as designed below, and the nested
+> `sdata/dist-nix` flake marked superseded. See "Sketch" for the per-item
+> status and "Open questions" for what the implementation resolved.
 
 ## Goal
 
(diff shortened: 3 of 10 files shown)


XephyLon (commented):
Reviewed at 0ec1c0179 (11 commits, 8 files, rebased onto today's main at 50abb16c2 — 0 merges, `rebaseable: true`, 0 conflicts against a dry merge). CI green (test + both receipts). What I could not do: run `nix flake check` / `nix build` — there is no nix on this machine, so the "Validated so far" section is taken on the doc's word; it predates the rebase, and the 11 commits main gained since touch nothing the flake reads.

## Should fix before merge

**1. The quickshell input points at a different quickshell than the one the shell is built and tested against.** `flake.lock` pins `quickshell` at 9f59d0a05 — the mirror's master as of today (2026-08-26, "wayland/lock: add warning on live session lock destruction", past v0.3.1). Every other path pins e649d247 (2026-07-21): `sdata/dist-arch/immaterial-impulse-quickshell-git/PKGBUILD` `_commit`, and the prebuilt the installer's fast path ships. The shell has already had to handle one ScriptModel representation change between quickshell revisions (`TaskList.qml`'s `resolveIndex`); a flake that floats on master hands NixOS users whatever quickshell merged last night. Unpinning the *URL* was right (Sketch item 4); the *lock* should sit on the PKGBUILD's commit, and `flake.nix`'s comment should name the PKGBUILD as the source of truth for `nix flake update quickshell`, or the two drift the way the Gentoo ebuild already has (7511545, March).

**2. The shell the module installs cannot generate colours, and nothing in the PR says so.** Eight scripts under `scripts/` — `colors/generate_colors_material.py` (the whole wallpaper → palette path), `colors/switchwall.sh`, `colors/scheme_preview.py`, `background/subject_mask.py`, `hyprland/hyprconfigurator.py` and the three `*-venv.sh` wrappers — run inside the venv `sdata/lib/package-installers.sh` creates (`uv venv … -p 3.12`, `sdata/uv/requirements.txt`) and find it through `IMMATERIAL_IMPULSE_VIRTUAL_ENV`. The module ships the tree, `dontPatchShebangs` keeps that exact `env -S … source $VENV` shebang intact — and provides neither the venv nor the variable, so the first wallpaper change fails. The doc's "scripts/ shells out to a lot of binaries" open question is about PATH; this is a separate, larger hole, and it is the one a NixOS user hits first. Two acceptable answers for this pass: name it explicitly in "Not yet validated"/"Open questions" as the first blocker of a usable shell, or close it — `python312.withPackages` from `sdata/uv/requirements.txt` plus `home.sessionVariables.IMMATERIAL_IMPULSE_VIRTUAL_ENV`, which is a small addition to a module that already manages the environment. Leaving it unnamed is the problem, not leaving it unfixed.

**3. No user-facing pointer to the flake.** `README.md` has no "Nix" anywhere; the only references are the proposal doc and the `sdata/dist-nix/README.md` "superseded" note — which a NixOS user only finds by already knowing where the old thing was. `Docs: updated` cites the proposal only. A root flake is an install path; README's install section should name `github:XephyLon/immaterial-impulse` and the module option, and say what does not work yet (item 2, Wallpaper Engine).

## Should fix, smaller

**4. `hm-module.nix`'s matugen output lists are a hand copy of `dots/.config/matugen/config.toml`.** In sync today (6 config + 5 state = the 11 `output_path` entries, checked), and nothing keeps them so — the next output added to `config.toml` is one the module silently stops warning about. A five-line contract in `tests/` (parse the TOML, compare against the two lists in the module) makes the drift red; it is the shape the rest of the suite uses.

**5. `nix flake check` in CI is the doc's own item 5 and stays a TODO.** With no nix on the maintainer's machine either, hand runs are the only thing guarding the flake, and they were last done before the rebase. `cachix/install-nix-action` on the Ubuntu runner makes it a 10-line job; until it exists, every merge to main can break the flake with no signal.

## Nits

- `docs/proposals/nixos-flake.md` header: "implemented on `feat/nixos-flake`" — the branch is `proposal/nixos-flake`.
- The doc cites the tree it checked as bare `4b43790` — it resolves, but a subject beside it is the house form and survives a rewrite.
- `.gitignore` has no `result` entry; `nix build` leaves one at the root.
- `nix/package.nix` uses `lib.cleanSource`, which does not drop `__pycache__`/`.pytest_cache` — a local build after a suite run ships them. `lib.cleanSourceWith` with a filter, or `lib.fileset`.
- Body receipt: `Changelog: not user-visible — entries ride the next catch-all release PR, per the release workflow`. Moot here (nothing under `dots/` changes, so no receipt is asked for), but CONTRIBUTING §Keep CHANGELOG.md fed describes the opposite workflow; paste this line into a PR that does touch `dots/` and it is the receipt-without-a-changelog shape the section warns about.

## Verified OK

- The premise: nothing in today's tree writes into `dots/.config/quickshell/imi` — no `setText`/`writeAdapter` targets `shellPath`/`scriptPath`/`assetsPath`, no script writes beside itself; a read-only store path holds.
- Root `VERSION` is a git symlink to the shell's file (mode 120000, repointed at the rename in cdb2d14da); `builtins.readFile ./VERSION` resolves it, so `version` is `0.31.1`, not the path.
- `tests/imports` symlink farm is relative and survives `cp -R` into the store.
- `quickshell.nix`'s `qsPackage ? quickshell.packages.x86_64-linux.default` — the nested flake's call still resolves through the lazy default; the root flake passes the per-system package.
- Seed-once activation is idempotent (`-e` guards, `run` for dry-run), and the tier-3 warnings cover exactly the 11 outputs.

**Verdict: not mergeable as-is on 1–3** — 1 is a one-line lock move, 2 is a paragraph (or a small module addition), 3 is a README section. 4–5 can follow. Re-review on request.

XephyLon:
Addressed 1–3 and the nits in 03618426..b622d28b; re-review when ready.

**1 — lock re-pinned.** `flake.lock`'s quickshell sits on the PKGBUILD's `_commit` (e649d247) now, and `flake.nix`'s input comment names the PKGBUILD as the source of truth and warns that a bare `nix flake update quickshell` lands on the mirror's master. `nix flake check` and `nix build .#immaterial-impulse` re-run green on the new lock (rootless nix-portable, same rig as before).

**2 — named, not fixed.** "Validated so far" now carries the blocker in full: the eight venv scripts, `IMMATERIAL_IMPULSE_VIRTUAL_ENV`, and why the first wallpaper change fails. I took the doc option rather than the module addition deliberately: `python312.withPackages` from `sdata/uv/requirements.txt` is not a small addition — the lock pins `materialyoucolor`, `material-color-utilities` and `kde-material-you-colors`, none of which nixpkgs' python set carries, so closing it means packaging those (or shipping the venv another way) and is its own follow-up. The PATH open question now points at this as the separate, larger hole.

**3 — README section added.** "NixOS (experimental)" under Install / update: the flake input line, the module import, the enable option, what the module does, and what does not work yet (colour generation, Wallpaper Engine), linking the proposal doc.

**Nits:** branch name in the doc header fixed; `4b437909 ("feat(background): tell the renderer when its output is covered")`; `/result` ignored; `cleanSourceWith` filter drops `__pycache__`/`.pytest_cache` — proven live, the tree carries `tests/__pycache__` and the store output does not; the PR body's Changelog line reworded to a reason that stands alone (it is moot here, as you say — nothing under `dots/`).

**4 and 5 left open** per the review ("can follow"): the matugen-output drift-guard test and the CI `nix flake check` job. Both are named in the doc's open list.

PR #76, opened 2026-08-02
fix(input): Super+Alt+Space no longer switches keyboard layout (#69)

### LINKED ISSUE #69: Super+Alt+Space (float toggle) also fires the keyboard-layout switch, and the layout never actually changes
Reported by the user.

## Symptoms

- Pressing **Super+Alt+Space** to release a window from the grid also triggers a keyboard-layout change.
- Pressing **Super+Space** afterwards does not change it back.
- The indicator shows the layout as changed, but typing continues in the same language regardless.

## Two findings from a first look

**1. The layout list has one entry, so switching cannot affect typing.**
`dots/.config/hypr/hyprland/general.lua:257` sets `kb_layout = "us"`. `hyprctl switchxkblayout all next` therefore cycles a single-element list — the indicator updates state, but the effective layout is unchanged. This matches "it shows that the layout has changed but it still types in the same language", and also explains why Super+Space does not put it back: there is nothing to put back.

**2. The float-toggle bind is colliding with the layout bind.**
- `keybinds.lua:42` — `hl.bind("SUPER + Space", hl.dsp.exec_cmd("hyprctl switchxkblayout all next"))`
- `keybinds.lua:212` — `hl.bind("SUPER + ALT + Space", hl.dsp.window.float({ action = "toggle" }))`

The comment at `keybinds.lua:39-41` says the layout switch was deliberately made a *compositor* bind rather than an xkb `grp:` option, because xkb toggles fire on Super+Space even with extra modifiers held. The report suggests the compositor bind has the same problem: Super+Alt+Space still matches the Super+Space bind.

## Worth checking

- Whether the layout switch should be a no-op (or hidden) when only one layout is configured, rather than showing an indicator change.
- Whether the Super+Space bind needs to be made exclusive so a superset of its modifiers does not match it.
- Whether the indicator reads its state from the compositor or tracks it independently — if the latter, it can drift from reality.

XephyLon:
Fixes #69.

## Problem
Super+Alt+Space (window float toggle) also fired a keyboard-layout change, and Super+Space afterwards wouldn't reverse it — the indicator showed a change but typing stayed in the same language.

## Root cause
Not the keybinds. Hyprland compositor binds match modifiers exactly, so `SUPER, Space` never fires under `SUPER+ALT`. The real culprit is a **leftover xkb `grp:win_space_toggle`** in the shipped default `config.json`. Commit `ff3aaff` moved layout switching from the loosely-matching xkb `grp:` option to a compositor bind, but missed the shipped default that the installer seeds. So fresh installs still applied `grp:win_space_toggle`, which:
- matches Super+Space loosely (fires even with Alt held) → accidental switch on the float toggle, and
- on plain Super+Space, ran *both* the xkb toggle and the compositor `switchxkblayout all next` — two advances over the two-entry `us, eg` list net back to the start, so typing never changed.

## Fix
`dots/.config/quickshell/imi/defaults/config.json`: `kbOptions: "grp:win_space_toggle"` → `""`. Super+Space is then handled solely by the exact-matching compositor bind, cycling the real `us, eg` list and reversing cleanly. The misleading-indicator-on-single-layout case is already guarded in `services/HyprlandXkb.qml`.

## Testing
- `config.json` validated as JSON.
- Traced the full path: default config → installer seed → `HyprlandConfig.setMany("input:kb_options", …)` → Hyprland runtime; confirmed `grp:win_space_toggle` appears nowhere else.

## Note for review
Existing installs are **not** auto-fixed — `seed_default_config` never overwrites an existing `~/.config/immaterial-impulse/config.json`. Affected users must clear the option in Settings → Hyprland → Input (or via a future one-time migration).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,10 @@ own repo; the installer pins which revision it builds.
   layout. The shipped default config still carried `input.kbOptions =
   grp:win_space_toggle` even after the layout switch was moved to a compositor
   bind — xkb `grp:` toggles match modifiers loosely, so Super+Space fired with
-  Alt held too. Fresh installs now seed `kbOptions = ""`; Super+Space runs only
-  the exact-matching `switchxkblayout all next` bind, which cycles the real
+  Alt held too. Fresh installs now seed `kbOptions = ""`, and a one-time
+  migration clears a stale `grp:win_space_toggle` from existing configs on the
+  next shell start, so both are fixed. Super+Space then runs only the
+  exact-matching `switchxkblayout all next` bind, which cycles the real
   `us, eg` layout list and reverses cleanly.
 
 ## [0.11.0] — 2026-08-02
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,15 @@ own repo; the installer pins which revision it builds.
 
 ## [Unreleased]
 
+### Fixed
+- Super+Alt+Space (window float toggle) no longer also switches the keyboard
+  layout. The shipped default config still carried `input.kbOptions =
+  grp:win_space_toggle` even after the layout switch was moved to a compositor
+  bind — xkb `grp:` toggles match modifiers loosely, so Super+Space fired with
+  Alt held too. Fresh installs now seed `kbOptions = ""`; Super+Space runs only
+  the exact-matching `switchxkblayout all next` bind, which cycles the real
+  `us, eg` layout list and reverses cleanly.
+
 ## [0.11.0] — 2026-08-02
 
 ### Changed
--- a/dots/.config/quickshell/imi/defaults/config.json
+++ b/dots/.config/quickshell/imi/defaults/config.json
@@ -626,7 +626,7 @@
     "input": {
       "followMouse": 1,
       "kbLayout": "us, eg",
-      "kbOptions": "grp:win_space_toggle",
+      "kbOptions": "",
       "numlock": true,
       "repeatDelay": 250,
       "repeatRate": 35,
--- a/dots/.config/quickshell/imi/modules/common/Config.qml
+++ b/dots/.config/quickshell/imi/modules/common/Config.qml
@@ -44,6 +44,23 @@ Singleton {
         obj[keys[keys.length - 1]] = convertedValue;
     }
 
+    // One-time migration (issue #69): older installs seeded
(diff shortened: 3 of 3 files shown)


XephyLon:
**Merge-order note: this conflicts with #11.**

`git merge-tree --write-tree HEAD gh/fix/69-layout-switch-modifier-collision` against `proposal/widgets-as-plugins`:

```
CONFLICT (content): Merge conflict in CHANGELOG.md
Auto-merging dots/.config/quickshell/imi/defaults/config.json
```

Both branches insert a new subsection immediately under `## [Unreleased]` — this PR adds `### Fixed`, #11 adds `### Added`. Trivial to resolve (keep both, `### Added` first per Keep a Changelog ordering), just needs doing by whichever lands second. `defaults/config.json` auto-merges fine: this PR touches `input.kbOptions`, #11 touches the `background.widgets.*` blocks.

The other four fix PRs (#72, #77, #78, #79) merge into #11 clean.

---

**Separately — this does not fix the reporter's machine.** `seed_default_config` (`sdata/subcmd-install/3.files.sh:66`) explicitly never overwrites an existing `~/.config/immaterial-impulse/config.json`, and `HyprlandConfig.qml:67` maps `input.kbOptions` → `input:kb_options` and re-applies it to Hyprland on every shell start. So an existing install keeps `grp:win_space_toggle` and keeps the bug. Verified on the reporter's live config:

```
"kbLayout": "us, eg",
"kbOptions": "grp:win_space_toggle"
```

The PR body says as much, so this is confirmation rather than a new finding — but it means merging this closes #69 only for fresh installs. A one-time migration is now cheap: `Config.qml` grew the marker pattern (`migratedDesktopWidgets` / `migratedDesktopWidgetOptions`) on #11's branch, so clearing a stale `grp:win_space_toggle` once behind its own marker would follow an established shape.

XephyLon:
Added the one-time migration (`f64ec24`): `Config.qml` now clears a stale `grp:win_space_toggle` on load behind a persisted `migratedKbOptionsGrpToggle` marker, so existing installs are fixed on next shell start — not just fresh ones. Self-contained (doesn't depend on #11's marker infra). CHANGELOG entry updated to match.

The #11 CHANGELOG merge-order note still stands: whichever of us lands second keeps both subsections under `## [Unreleased]` (`### Added` first).

PR #77, opened 2026-08-02
fix(background): wallpaper goes blank after a transition on GL 2.1 (#70)

XephyLon:
Fixes #70.

## Problem
Changing the wallpaper played the transition animation, then the desktop went **permanently blank** (reporter on Fedora, OpenGL 2.1-class GPU). Logs showed the transition fragment shader failing to compile, retrying forever.

## Root cause
The "melt" transition is the **Doom** screen-wipe. Qt's RHI cross-compiles the baked shader to the **GLSL ES 1.00 (`#version 100`)** profile on GL 2.1 backends, which forbids two constructs the Doom shader used:
- a global `const int _45[256]` lookup table → `error C7516` / `C7553`
- a bitwise `int & 255` → `error C7548` (needs `GL_EXT_gpu_shader4`)

So the `ShaderEffect` never compiled; because the plain wallpaper `Image` is hidden while a transition is configured, the desktop stayed blank.

## Fix
- **`Doom.frag` + `Doom.frag.qsb`** — replaced the `const int[256]` table with equivalent procedural value-noise (hashed lattice + smoothstep), normalized to the original `[1.0, 2.0]` offset range; replaced `& 255` with `mod(…, 256.0)`. Rebaked the `.qsb` (`qsb --glsl "100es,120,150" --hlsl 50 --msl 12`). *Only `.qsb` artifacts were previously committed — I reconstructed `Doom.frag` in the same `#version 440` form as `Peel.frag` so the tracked source now matches the artifact.*
- **`Background.qml`** — graceful fallback: on `ShaderEffect.Error`, settle the transition and reveal the plain wallpaper instead of blank output. Additive; render path is byte-identical when shaders compile.

## Testing
- Dumped the rebaked GLSL ES 1.00 variant: **zero** `const int` / `int[` / `[256]` / bitwise `&`; only ES-1.00-valid builtins remain — directly eliminating all three errors.
- Scanned every other transition `.qsb` (circle, circlePit, circleSelect, magic, Peel, transition, pixelate, stripes) — none had the offending constructs, so Doom was the only affected shader.
- `qmllint` clean on `Background.qml` (module-resolution warnings only).

## Notes for review
- The melt is now a close **approximation** of the Doom column pattern (timing/push-range/character preserved, not the byte-exact 256-entry curve).
- Not live-verified from an isolated worktree — a real switch on the primary checkout (ideally on the reporter's GL 2.1) is the final check, especially the fallback's visibility bindings.

PR #78, opened 2026-08-02
fix(bar): closing hover popup no longer paints over its neighbour (#71)

### LINKED ISSUE #71: entry of tooltip looks bad if there was one active before
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://ii.clsty.link/en/ii-qs/04troubleshooting/) and [Usage](https://ii.clsty.link/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://ii.clsty.link/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

...

### Step 3. Describe the issue

When you have for example the clock and the weather widget besides each other and you hover over the first one, the tooltip appears cleanly. But when you then move your cursor to the second widget, the entry of the new tooltip / exit of the old tooltip seem to clip / overlay in an ugly way: The new widget appears in front of the old one, but then the old one starts the exit animation and for the time of that animation, it is on top of the new tooltip.

### Reminder

- [x] I agree that it's usually impossible for others to help me without my logs.

XephyLon:
Fixes #71.

## Problem
With two adjacent bar widgets (e.g. clock + weather), hovering the first then sweeping to the second made the **old tooltip's exit** render on top of the **new tooltip's entry** for the duration of the close animation.

## Root cause
The bar hover popups are each their own **layer-shell surface** (`StyledPopup.qml` → `PanelWindow`, `WlrLayer.Overlay`), not sibling `Item`s. Each has a 180 ms hover-hold grace period so you can move onto a popup without it vanishing. Moving clock→weather opens the new popup immediately while the old one lingers for up to 180 ms; both are ~340 px wide and centred under adjacent widgets, so they overlap — and same-layer wlr surfaces resolve overlap by **stacking order, not activeness**, so the lingering one can cover the new one. A `z` fix doesn't apply (independent Wayland surfaces).

## Fix
Temporal, not stacking. New shared slot `GlobalStates.activeBarPopup`:
- When a popup's target widget is hovered, it claims the slot.
- A `Connections` on the slot makes any *other* popup that is only lingering (`hoverHeld && !targetHovered && !popupHovered`, and not pinned) stop its timer and close immediately — in the same event-loop turn, before the next frame — so the two are never mapped simultaneously.

Lives in the one shared `StyledPopup`, so every bar hover popup benefits. Intra-popup travel (widget → its own popup) and pinned/interactive popups are unaffected.

## Testing
- Repo suite `dots/.config/quickshell/imi/tests/run_tests.sh`: **200 passed, 0 failed** (includes the QML import linter + shared-widget contract tests).
- Traced both repro chains (ClockWidget/WeatherBar → …Popup → StyledPopup) to confirm they share the widget.

## Note for review
Couldn't hot-reload the live shell from the worktree — please do one live confirmation: hover clock, sweep to weather and back (old popup no longer flashes on top), and confirm a pinned popup still stays open when a neighbour is hovered.

### DIFF
--- a/dots/.config/quickshell/imi/GlobalStates.qml
+++ b/dots/.config/quickshell/imi/GlobalStates.qml
@@ -46,6 +46,11 @@ Singleton {
     property real desktopMenuX: 0
     property real desktopMenuY: 0
     property string wallpaperSelectorTarget: "wallpaper"
+    // The bar hover popup (StyledPopup) whose target widget is currently
+    // hovered. Adjacent bar popups are separate layer-shell surfaces, so a
+    // lingering one can paint over a newly opened neighbour; each popup watches
+    // this so the previous one closes at once instead of overlapping the new one.
+    property var activeBarPopup: null
     property bool dropShelfOpen: false
     property real dropShelfX: 0
     property real dropShelfY: 0
--- a/dots/.config/quickshell/imi/modules/common/widgets/StyledPopup.qml
+++ b/dots/.config/quickshell/imi/modules/common/widgets/StyledPopup.qml
@@ -1,3 +1,4 @@
+import qs
 import qs.modules.common
 import qs.modules.common.widgets
 import qs.modules.common.functions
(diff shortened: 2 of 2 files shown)


XephyLon:

XephyLon:
Added a CHANGELOG `### Fixed` entry (`183107b`) so the fix shows up in `setup exp-update`'s What's New.

PR #79, opened 2026-08-02
fix(colors): sync kitty colors with matugen (#75)

### LINKED ISSUE #75: Terminal colors do not sync with matugen
Changing the wallpaper re-themes the shell, GTK apps, KDE apps, fuzzel, hyprlock, cava and tmux — but the terminal keeps whatever colors it shipped with.

## Cause

`dots/.config/matugen/config.toml` declares templates for:

```
m3colors, hyprland, hyprlock, fuzzel, gtk3, gtk4, kde_colors, wallpaper, cava, tmux
```

There is **no terminal emulator template of any kind** — no kitty, foot, alacritty, wezterm, ghostty. `dots/.config/matugen/templates/` has no terminal directory either, and neither `scripts/colors/apply_matugen_app_themes.py` nor `install_matugen_app_themes.sh` mentions a terminal.

So this is a missing template, not a broken one. `tmux` being present makes it read as though terminal theming is covered when only the multiplexer *inside* the terminal is.

## Scope

Worth deciding which emulators to support before writing templates. Each needs a different reload mechanism, and that is usually the harder half:

- **kitty** — `kitty @ set-colors` over its remote-control socket, which requires `allow_remote_control` to be on. Falls back to needing a restart.
- **foot** — reads `~/.config/foot/foot.ini` at launch; no live reload for colors.
- **alacritty** — live-reloads its TOML on write.
- **ghostty** / **wezterm** — each has its own story.

`scripts/colors/switchwall.sh` is where the post-matugen hooks live and is the natural place to trigger whatever reload each emulator needs.

## Related

`sdata/deps-info.md` and the per-distro dep lists would need the emulator names if any reload path grows a binary dependency.

XephyLon:
Fixes #75.

## Problem
The wallpaper palette re-themed shell/GTK/KDE/fuzzel/hyprlock/cava/tmux, but kitty had no **matugen-native** template — it was the one GUI app with no `config.toml` entry.

## Finding
Kitty is *already* recolored on every wallpaper switch by a separate, non-matugen path: `switchwall.sh` → `applycolor.sh` → `apply_kitty()` regenerates `kitty-theme.conf` from the tuned `term0..15` terminal palette and live-reloads via `kill -SIGUSR1 $(pidof kitty)`. This PR adds a matugen-native source **in addition** to that (decision: keep both).

## Fix
- **`matugen/templates/kitty/colors.conf`** (new) — renders bg/fg, cursor(+text), selection, and `color0..15` from Material roles.
- **`matugen/config.toml`** — `[templates.kitty]` → `~/.config/kitty/colors-matugen.conf` (dedicated file, fuzzel pattern, so user kitty settings are never clobbered).
- **`kitty.conf`** — `include colors-matugen.conf` ordered **after** the existing theme include, so the matugen palette wins the 16 ANSI slots + bg/fg/cursor while the older file keeps its `color232-255` starship overrides and background-image block.
- **`apply_matugen_app_themes.py`** — `reload_kitty()`: guarded `kitty @ set-colors --all --configured <file>`, a no-op when kitty is absent / not running / remote control off (this repo ships RC off; running windows are recolored by `applycolor.sh`'s SIGUSR1 in the same switch).

## Color-slot mapping
Material 3 has only three accent hues + error, so ANSI names can't be honored literally: `color1/9 = error` (red stays an alert), `0/8 = surface_container_lowest/surface_variant`, `7/15 = on_surface_variant/on_surface` (neutrals), `2–6`/`10–14` cycle primary/secondary/tertiary and their fixed/dim variants.

## Testing
- `config.toml` parses with `tomllib` (11 templates, kitty present); `py_compile` clean on the edited script.
- Rendered the template with dummy hex (all 19 placeholders substituted) and had **kitty itself parse it through the include chain** (`kitty +runpy load_config`) — all colors resolve, no errors.
- No live wallpaper switch run (would mutate the system).


XephyLon:

XephyLon:
Added a CHANGELOG entry (`272fce1`). Filed under `### Added` rather than `### Fixed` since it's a new matugen template/capability layered on top of the existing terminal palette pass — say the word if you'd rather it read as Fixed to match the issue.

PR #85, opened 2026-08-02
fix(panels): blur only the panel body, not the drop shadow (#82)

XephyLon:
Fixes #82.

> **Approach replaced** (branch force-pushed early on): the first version raised the sidebars' layerrule `ignore_alpha` to 0.6. Live testing showed the frost persisted on busy wallpapers; researching how DankMaterialShell renders crisp panel shadows led to a fundamentally better mechanism, which was then extended to every panel family at maintainer request.

## Problem
The sidebars — and equally the bar, vertical bar and dock — draw translucent drop shadows (`StyledRectangularShadow`) and rounding/elevation margins in the same layer surface as their body. Hyprland's catch-all `quickshell:.*` layerrule blurs the **whole surface**, shadow and margins included, leaving a frosted band along panel edges (obvious on busy wallpapers, subtle on dark ones).

## Fix (DMS-style blur region)
Scope the compositor blur to the painted body shapes only, via `ext_background_effect_manager_v1` (Hyprland ≥ 0.51; exposed by Quickshell as `BackgroundEffect.blurRegion`):

- **`modules/common/widgets/WindowBlurRegion.qml`** (new) — publishes a blur `Region` (item-tracked, radius-matched) on a panel window. Declarative binding plus a short settle-republish after map/resize, mirroring DankMaterialShell's `WindowBlur` "kick": a region committed mid-configure can be dropped compositor-side. Simple `regionItem` API for single-rect bodies; an overridable `region` property lets multi-rect bodies compose child `Region`s into a union.
- **Sidebars** (`ffdab66`) — region = the body rect each window's input `mask` already tracks (`SidebarRightContent` exposes its nested body via `backgroundItem`).
- **Bar** (`09043d8`) — composed region over the *painted* shapes: main background (Hug/Float styles) + center-only pill, each gated on the exact condition that paints it. A blur region is a plain rect, so unpainted/transparent shapes must stay out — covering them would frost bare wallpaper. `RoundCorner` decorators and islands/material pills are deliberately uncovered: opaque by default, they merely read as unblurred translucency under transparency mode.
- **Vertical bar + dock** (`d1e1dce`) — vertical bar mirrors the bar exactly (same exposed contract); the dock is a single rect gated on `dock.showBackground`.
- **`rules.lua`** — layerrule `blur = false` for all five namespaces (`sidebarLeft/Right`, `bar`, `verticalBar`, `dock`), so the region is the sole blur source.

[...]

PR #86, opened 2026-08-02
feat(bar): adjustable bar background opacity (#83)

### LINKED ISSUE #83: Support for (partly) transparent bar
### What would you like to be added?

I think it would be awesome to have a setting to set the bar opacity.

### How will it help?

It would make the bar and widgets feel more integrated on setups that have transparent / blurred windows.

### Extra info

It is possible that I just missed a setting. I think this is also possible to do through configuration in the ~/.config/hypr/custom/rules.lua file, but I would be happy to hear your thoughts on the idea of a setting in the shell.

XephyLon:
Closes #83.

## What
Adds a first-class **`bar.backgroundOpacity`** setting (real, 0–1, default **1.0 = fully opaque**, so the out-of-box look is unchanged) with a Settings slider, for transparent/blurred setups.

## How
- `Config.qml` — `property real backgroundOpacity: 1.0` under `bar` (missing key falls back to the default; no migration needed).
- `Appearance.qml` — a derived `colBarBackground = transparentize(colLayer0, 1 - backgroundOpacity)`. At `1.0` it equals `colLayer0`. It *multiplies* `colLayer0` (which already carries the global `appearance.transparency` alpha) rather than setting an absolute alpha, so the two compose instead of fighting.
- `Bar.qml` / `BarContent.qml` / `VerticalBar.qml` / `VerticalBarContent.qml` — 18 bar-background / pill / hug-corner fills swapped `colLayer0 → colBarBackground` (horizontal + vertical bars, all four corner styles). `colLayer0Border` deliberately untouched.
- `settings/pages/BarConfig.qml` — a `ConfigSlider` (0–1) after "Bar shadow", gated `enabled: bar.showBackground`.
- `tests/tst_config.qml` — asserts the default stays `1.0`, so an accidental non-opaque default (which would silently thin the whole bar) fails the suite.

## Design decisions
- **Own `bar.` setting, not under `appearance.transparency`** — the global block thins *every* surface; the reporter wants the bar specifically. Matches sibling precedent (`dropShelf.backgroundOpacity`, `terminal.background.opacity`).
- **Independent of `transparency.enable`** but composes with it.
- **Scope = the bar's own background chrome only** (bar fill, center/material pills, hug corners). Widget pill fills stay on `colLayer1` so contents stay readable; a fully see-through bar is already available via Group style → none.
- **Blur:** at default `1.0` nothing changes. ~~At low opacity the surface falls below the bar namespace's compositor `ignore_alpha` blur threshold and reads as plain (unblurred) transparency.~~ **Updated:** with #85 merged (body-scoped blur regions), a lowered-opacity bar body gets a proper **frosted-glass** backdrop blur — verified live on the merged tree. Without #85, the pre-existing whole-surface blur behavior applies. Either way this PR is safe to merge in any order relative to #85 (auto-merge verified clean both ways).

## Testing
- `dots/.config/quickshell/imi/tests/run_tests.sh`: 200 passed (import lint, spacing lint, shared-widget contracts, new config-default assertion).
- **Verified live**: slider works across corner styles; bar/pills go translucent while icons/text stay crisp.

_CHANGELOG/VERSION intentionally omitted — release notes land in the catch-all release PR._

### DIFF
--- a/dots/.config/quickshell/imi/modules/common/Appearance.qml
+++ b/dots/.config/quickshell/imi/modules/common/Appearance.qml
@@ -126,6 +126,12 @@ Singleton {
         property color colLayer0Hover: ColorUtils.transparentize(ColorUtils.mix(colLayer0, colOnLayer0, 0.9, root.contentTransparency))
         property color colLayer0Active: ColorUtils.transparentize(ColorUtils.mix(colLayer0, colOnLayer0, 0.8, root.contentTransparency))
         property color colLayer0Border: ColorUtils.mix(root.m3colors.m3outlineVariant, colLayer0, 0.4)
+        // The bar's own background chrome (bar/pill fills, hug corners) thins
+        // colLayer0 by the user's bar opacity (Config.options.bar.backgroundOpacity,
+        // 1 = fully opaque = unchanged). Kept separate from colLayer0 so only the
+        // bar goes translucent, composing with — not overriding — the global
+        // transparency setting colLayer0 already carries.
+        property color colBarBackground: ColorUtils.transparentize(colLayer0, 1 - (Config?.options.bar.backgroundOpacity ?? 1))
         // Layer 1
         property color colLayer1Base: m3colors.m3surfaceContainerLow
         property color colLayer1: ColorUtils.solveOverlayColor(colLayer0Base, colLayer1Base, 1 - root.contentTransparency);
--- a/dots/.config/quickshell/imi/modules/common/Config.qml
+++ b/dots/.config/quickshell/imi/modules/common/Config.qml
@@ -878,6 +878,12 @@ Singleton {
                 property string borderless: "pills"
                 property string topLeftIcon: "spark" // Options: "distro" or any icon name in ~/.config/quickshell/imi/assets/icons
                 property bool showBackground: true
+                // Opacity of the bar's background chrome (bar/pill fills, hug
+                // corners), 0-1. 1 = fully opaque (unchanged). Multiplies the
+                // global appearance.transparency alpha rather than replacing it,
+                // so the two compose. Below the compositor's per-namespace
+                // ignore_alpha blur threshold it reads as plain transparency.
+                property real backgroundOpacity: 1.0
                 property bool verbose: true
                 property bool vertical: false
                 property JsonObject resources: JsonObject {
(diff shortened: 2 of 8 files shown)


XephyLon:
**Heads up: this now fails the suite on `main`.** Nothing wrong with the change itself — `main` moved underneath it.

`54190291f`..`1e435b5cb` landed a fix for a bug where **opening a Settings page rewrote config values it merely displayed**: QQC2 bounds `value` into `[from, to]` at component completion, and every ranged control wrote back from `onValueChanged`, so building the page clamped and persisted anything out of range (a configured `osd.timeout: 4321` became `3000`, no user, no interaction). `ConfigSlider` and `ConfigSpinBox` now raise `valueModified(real newValue)` instead, and all 69 existing call sites were converted.

Merged into current `main`, this PR trips the guard that keeps it from coming back:

```
FAIL: test_no_ranged_control_writes_from_on_value_changed
AssertionError: Lists differ:
  ['modules/imi/settings/pages/BarConfig.qml:292 (ConfigSlider)'] != []
```

One-line fix — same shape as the other sliders on this page (e.g. `BarConfig.qml:346`):

```qml
-                    onValueChanged: {
-                        Config.options.bar.backgroundOpacity = value;
-                    }
+                    onValueModified: {
+                        Config.options.bar.backgroundOpacity = newValue;
+                    }
```

Note `newValue` rather than `value` — the signal carries the value the user landed on. Worth taking, beyond the failing test: `ConfigSlider`'s `StyledSlider` has a `Behavior on value`, so one external change used to emit ~25 `valueChanged` events, i.e. 25 config writes of intermediate animation frames.

Everything else checks out. Verified locally against `1e435b5cb`: merges clean, and with that one line changed the full suite is green (**250 assertions**, up from the 200 you tested against). **#85 needs no changes** — it merges clean and passes as-is.

XephyLon:
Taken — thanks for the exact patch. Rather than stacking a fixup on the stale base (where `valueModified` doesn't exist yet, leaving the branch broken standalone), I **rebased onto current `main` (`1e435b5`) and folded the one-line change into the feature commit** (`1a8cc62` → `2cd85ce`, same diff otherwise). The slider now uses `onValueModified` / `newValue`, matching the converted call sites around it — and good point on the `Behavior on value` write-amplification; that alone justified the switch.

Verified on the rebased branch: **250 assertions, 0 failed**, including `test_no_ranged_control_writes_from_on_value_changed`.

PR #105, opened 2026-08-04
Make the WE prebuilt installer tests actually run, and fix the dead skip path they were written to catch

### LINKED ISSUE #91: test_wallpaperengine_prebuilt.py never runs: pytest-style file invoked as a plain script, so the suite is always green
`dots/.config/quickshell/imi/tests/run_tests.sh:314` runs the Wallpaper Engine prebuilt installer tests like every other Python test in the suite:

```sh
echo "Running Wallpaper Engine prebuilt installer tests..."
if ! python3 "$SCRIPT_DIR/test_wallpaperengine_prebuilt.py"; then
```

But `tests/test_wallpaperengine_prebuilt.py` is written for **pytest**, not as a script: its seven checks are module-level `def test_*(tmp_path)` functions taking pytest's `tmp_path` fixture, and the file has no `if __name__ == "__main__"` block and no calls at the bottom. Running it with plain `python3` defines the functions, executes nothing, and exits 0.

Verified: `python3 test_wallpaperengine_prebuilt.py` → `RC=0`, no output. It is the **only** `test_*.py` in `tests/` without a `__main__` runner, and the only one that uses `tmp_path`; every other file in the directory self-runs.

### What breaks

- The step is permanently green regardless of the installer's behaviour.
- `.github/workflows/tests.yml` runs `run_tests.sh` on every push to `main` and every PR, and installs only Qt6 packages — no pytest — so nothing anywhere in CI executes these tests.
- `test_prebuilt_happy_path` asserts `r.returncode == 0` and would have caught #90 (the prebuilt path aborts with `lib_dir: unbound variable` and exits 1) the moment it was written. The one path that cannot be exercised in production is also the one path with no working test.

### Fix shape

Either add a `__main__` block that calls each test with a `tempfile.mkdtemp()` path (matching the convention the rest of the directory already uses), or install pytest in CI and invoke it via `python3 -m pytest`. The former keeps `run_tests.sh` dependency-free.

**Severity: high.** Not a user-visible bug, but it silently voids the only coverage of the install path that runs on other people's machines, and it demonstrably let a critical defect ship.

XephyLon:
## Describe your changes

Closes #91.

`tests/test_wallpaperengine_prebuilt.py` was written for pytest — seven module-level `def test_*(tmp_path)` functions, no `__main__` block — but `run_tests.sh` invokes every Python check as `python3 <file>`. It defined its functions, called none of them, and exited 0. pytest is not installed in CI, so nothing anywhere has ever executed these assertions.

Making them run immediately surfaced a real bug, which is the second commit here.

**`up_to_date` could never return 0.** `write_stamp` records three fields (`ref`, binary, lib dir); `up_to_date` read them into two variables. bash's `read` assigns every remaining field to the last name, so `bin` held `"<binary> <libdir>"` as one string — never a path to an executable, so `[[ -x "$bin" ]]` could not pass. The skip was dead: **every re-run of the installer went through a full `try_prebuilt` or source build**, which for anyone without a prebuilt release is the 5–40 minute recompile the stamp exists to avoid, on every update, including ones that did not move `WE_REF`. It hid because the code on the far side of the gate is correct — the main block already reads three fields — so it never looked like a broken conditional, only like a slow installer.

Three commits, kept separate: the installer fix, the test conversion, and the CI dependency.

### What the tests now pin

Nine cases, up from seven. Two are new: that each stamp field is recoverable *on its own* (the property that actually broke — "the stamp has three fields" would not have caught it, since the writer was always correct), and that taking the skip still rewrites the wrapper (deliberate behaviour with its own regression history — a wrapper-only fix could not reach an existing install, which is how an `LD_LIBRARY_PATH` leak survived an update).

`test_stale_stamp_ref_does_not_skip` was rewritten because it did not test what it claimed — see the PR comment for how that was found.

### Verification

Every assertion was proven by mutation: nine separate breakages of `4.wallpaperengine.sh`, each confirmed to turn the file red, each reverted. Full suite green at **260 passed, 0 failed** before each commit. Details in the comment below.

Ready to review. Not merged, nothing tagged, no release cut, `WE_REF` and `SDDM_REF` untouched.

Two things worth a maintainer's eye, both called out in the comment: the installer fix is a behaviour change beyond the issue's stated scope (it was required to make the tests pass), and `CONTRIBUTING.md` points at a `tests/README.md` that does not exist in the repo.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,7 +17,17 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y qt6-declarative-dev qml6-module-qttest qml6-module-qtquick qml6-module-qtqml qml6-module-qtquick-controls qml6-module-qtquick-layouts qml6-module-qtqml-workerscript qml6-module-qtquick-window
-      
+
+      # zstd builds the fixture release tarballs in
+      # test_wallpaperengine_prebuilt.py and extracts them in the installer under
+      # test. It is preinstalled on ubuntu-latest, so this is belt-and-braces -
+      # but that file spent its whole life reporting success without running
+      # anything, and it now refuses to pass when the tool is absent rather than
+      # quietly covering nothing. Naming the dependency keeps a future runner
+      # image change from turning that refusal into a mystery red build.
+      - name: Install release-fixture tooling
+        run: sudo apt-get install -y zstd
+
       - name: Run Tests
         env:
           QT_QPA_PLATFORM: offscreen
--- a/dots/.config/quickshell/imi/tests/test_wallpaperengine_prebuilt.py
+++ b/dots/.config/quickshell/imi/tests/test_wallpaperengine_prebuilt.py
@@ -207,6 +207,27 @@ class WallpaperEnginePrebuiltTest(unittest.TestCase):
         self.assertTrue(pathlib.Path(lib).is_dir(),
                         f"stamp field 3 is not a directory on its own: {lib!r}")
 
+    def test_a_stamp_naming_a_missing_binary_does_not_skip(self):
+        # up_to_date tests `-x` on the recorded binary because a matching ref
+        # does not prove the build output survived: a cleared cache dir, a
+        # pruned tmpfs or a manual rm all leave a current-looking stamp pointing
+        # at nothing. Skip on that and the wrapper is rewritten to exec a file
+        # that is not there, so the install reports success and the breakage
+        # surfaces later, at shell start. Deleting the `-x` line passed every
+        # other case in this file.
+        make_release(self.rel)
+        self.assertEqual(self.run_installer().returncode, 0)
+        binary = pathlib.Path(self.stamp.read_text().split()[1])
+        self.assertTrue(binary.exists(), "nothing recorded to delete")
+        binary.unlink()
+        # The fixture stays put, unlike the skip tests above: this run is
+        # *supposed* to reinstall, so it needs something to reinstall from.
+        r = self.run_installer()
+        self.assertEqual(r.returncode, 0, r.stderr)
+        self.assertNotIn("skipping", self.out(r),
+                         "stamp ref matched but its binary was gone, and the install skipped anyway")
+        self.assertTrue(binary.exists(), "the missing binary was not reinstalled")
+
     def test_skip_path_still_refreshes_the_wrapper(self):
         # The skip guards the expensive fetch/build, NOT the wrapper. A
         # wrapper-only fix has to reach an existing install, and once did not:
--- a/dots/.config/quickshell/imi/tests/test_wallpaperengine_prebuilt.py
+++ b/dots/.config/quickshell/imi/tests/test_wallpaperengine_prebuilt.py
@@ -107,6 +107,18 @@ class WallpaperEnginePrebuiltTest(unittest.TestCase):
         return subprocess.run(["bash", str(SH)], env=self.env(**extra),
                               capture_output=True, text=True, timeout=300)
 
+    def fake_pacman(self, qt_version):
+        # The host-Qt gate reads `pacman -Q qt6-base`. Off Arch there is no
+        # pacman, host_qt comes back empty and the whole comparison
+        # short-circuits - so without this stub that gate is unreachable and any
+        # test claiming to cover it would be passing vacuously.
+        bindir = self.tmp / "fakebin"
+        bindir.mkdir(exist_ok=True)
+        p = bindir / "pacman"
+        p.write_text(f'#!/bin/sh\n[ "$1" = "-Q" ] && echo "qt6-base {qt_version}"\n')
+        p.chmod(0o755)
+        return str(bindir) + os.pathsep + os.environ.get("PATH", "")
+
     @staticmethod
     def out(r):
         return (r.stdout + r.stderr).lower()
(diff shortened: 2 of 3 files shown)


XephyLon:
## What was wrong, confirmed before changing anything

`python3 test_wallpaperengine_prebuilt.py` → `RC=0`, no output. In the pre-change suite log the step is two adjacent lines:

```
Running Wallpaper Engine prebuilt installer tests...
Running keybind cheatsheet parser tests...
```

Banner, then the next step. Nine function definitions, zero calls, exit 0.

Six of the seven original tests passed once forced to run. **The seventh did not**, and it found a live bug.

## The bug the dead test was written to catch

`write_stamp` records three fields — ref, binary, lib dir. `up_to_date` read them into two names:

```sh
read -r ref bin < "$STAMP_FILE"   # bin == "/…/quickshell /…/lib"
```

bash's `read` puts every remaining field into the last variable, so `bin` held both paths as one string. That is never a path to an executable, so `[[ -x "$bin" ]]` could not pass and **`up_to_date` could never return 0**. Demonstrated directly:

```
read -r ref bin      -> bin='/cache/bin/quickshell /cache/lib'   -> -x FAILS
read -r ref bin lib  -> bin='/cache/bin/quickshell'              -> -x OK
```

The skip was dead. Every re-run went through `try_prebuilt` or a full `source_build` — for anyone without a prebuilt release, the 5–40 minute recompile the stamp exists to prevent, on every update, including ones that never moved `WE_REF`. It hid because the main block already does `read -r _stamp_ref _stamp_bin _stamp_lib`; the code past the gate was correct, so this never read as a broken conditional, only as a slow installer.

## What the tests now pin

Eleven cases, up from seven, `unittest.TestCase` + `unittest.main()` per CONTRIBUTING.md — no new dependency.

| test | pins |
|---|---|
| `test_prebuilt_happy_path` | valid release installs, wrapper lands in the prefix |
| `test_tamper_falls_back` | checksum mismatch declines **and installs nothing** |
| `test_arch_mismatch_falls_back` | manifest `arch` gate fires on its own, without help from the filename |
| `test_smoke_failure_falls_back` | a binary that fails `--version` is not installed |
| `test_host_qt_older_than_build_qt_falls_back` | `qt_min` > host qt6-base declines |
| `test_host_qt_new_enough_installs` | an adequate host Qt is still accepted |
| `test_second_run_skips_when_up_to_date` | the skip fires, and needs no release fixture to do it |
| `test_stamp_fields_are_individually_recoverable` | each stamp field is usable **alone** |
| `test_skip_path_still_refreshes_the_wrapper` | taking the skip still rewrites the wrapper |
| `test_force_rebuild_overrides_skip` | `WE_FORCE_REBUILD` beats a matching stamp |
| `test_stale_stamp_ref_does_not_skip` | a ref mismatch declines the skip |

Three deserve explanation.

**`test_stamp_fields_are_individually_recoverable`** asserts field 2 is executable and field 3 is a directory *on their own*. Asserting "the stamp has three fields" would not have caught this bug — the writer was always correct; the reader folded two fields into one.

**`test_skip_path_still_refreshes_the_wrapper`** pins deliberate behaviour with its own regression history (a wrapper-only fix could not reach an existing install, which is how an `LD_LIBRARY_PATH` leak survived an update and a restart). Nothing held it in place.

**`test_stale_stamp_ref_does_not_skip` was rewritten because it did not test what it claimed** — and I only know that because mutation 7 below initially *survived*. The old version wrote a stale stamp into a fresh prefix, so `up_to_date` bailed on the missing-`$PREFIX/bin/quickshell` check and never reached the ref comparison. Deleting that comparison from the script entirely left the test green. It now installs first and rewrites only the ref, so the comparison is the sole gate that can decline.

The `qt_min` gate had the same defect waiting: it reads `pacman -Q qt6-base`, so off Arch `host_qt` is empty and the comparison short-circuits. On ubuntu-latest a test written the obvious way would have passed without evaluating anything. Both Qt cases stub `pacman` onto `PATH` so the gate is genuinely reached.

## Mutations

Each one breaks the subject script, runs the file, and is reverted. Control: clean tree passes.

| # | mutation to `4.wallpaperengine.sh` | suite | killed by |
|---|---|---|---|
| 1 | `read -r ref bin lib` → `read -r ref bin` (restore the bug) | **FAILED** | `second_run_skips`, `skip_path_still_refreshes` |
| 2 | checksum gate → `if false` | **FAILED** | `tamper_falls_back` |
| 3 | manifest-arch gate line deleted | **FAILED** | `arch_mismatch_falls_back` |
| 4 | smoke-test gate → `if false` | **FAILED** | `smoke_failure_falls_back` |
| 5 | `install_wrapper "$qs_bin" "$lib"` → `:` | **FAILED** | `prebuilt_happy_path` +4 |
| 6 | `WE_FORCE_REBUILD` line deleted | **FAILED** | `force_rebuild_overrides_skip` |
| 7 | `[[ "$ref" == "$WE_REF" ]] \|\| return 1` deleted | **FAILED** | `stale_stamp_ref_does_not_skip` |
| 8 | `write_stamp` emits 2 fields | **FAILED** | `stamp_fields_are_individually_recoverable` +1 |
| 9 | `install_wrapper "$_stamp_bin" "$_stamp_lib"` → `:` | **FAILED** | `skip_path_still_refreshes_the_wrapper` |
| 10 | host-Qt gate → `if false` | **FAILED** | `host_qt_older_than_build_qt_falls_back` |
| 11 | host-Qt gate → `if true` | **FAILED** | `host_qt_new_enough_installs` +7 |

**All 11 fail. Tree verified clean afterwards.** The harness also aborts a mutation whose `sed` did not apply, so a no-op edit cannot be misread as a passing result.

Mutation 7 is the one worth noting: on the first run it **survived**, and that is how the vacuous stale-ref test was found. It was not visible by reading.

## Full suite

`cd dots/.config/quickshell/imi/tests && bash run_tests.sh`, run before each of the four commits:

```
Totals: 260 passed, 0 failed, 0 skipped, 0 blacklisted
All tests passed successfully!
EXIT=0
```

The WE step now prints 11 results where it printed nothing. (`Ran 11 tests` appears twice in the log — the second is `test_uninstall_login_shell.py`, which coincidentally also has 11. This file is invoked once, at `run_tests.sh:322`.)

## Decisions and things I could not do

**The installer fix goes beyond the issue's stated scope.** #91 asks only that the tests run. But `test_second_run_skips_when_up_to_date` cannot pass against a broken `up_to_date`, and the brief was to make them run *and pass*. Loosening the test to accommodate the bug would have recreated the exact failure being fixed. It is a separate first commit so it can be reverted independently.

**Test isolation changed.** `PREBUILT_ROOT` defaulted to the real `~/.cache/immaterial-impulse`, so running the suite wrote into the user's actual install cache, and one test's extracted tree could satisfy another's `-x` check. Every `WE_*` variable is now cleared and set per-test, so an exported `WE_FORCE_REBUILD` in a developer's shell cannot alter what these assert.

**A missing `zstd` is a hard error, not a skip.** The defect here was a check that quietly covered nothing; a skip on a missing tool is the same outcome in a different costume. `zstd` is also now named explicitly in `tests.yml` — it is preinstalled on ubuntu-latest, so that is belt-and-braces against a future runner image change.

**Not done, deliberately:** not merged, nothing tagged, no release cut, `WE_REF` and `SDDM_REF` untouched. No merge commits — four linear commits on the branch.

**No AGENT.md update.** CONTRIBUTING.md asks for one when the test suite changes, but AGENT.md contains no description of this file, the WE installer, or the stamp mechanism (`grep` for `wallpaperengine.sh`, `we-installed-ref`, `WE_REF`, `stamp` returns nothing), so there is nothing to bring back into sync. Inventing a section felt worse than leaving it. Flagging rather than deciding silently.

**`tests/README.md` does not exist.** CONTRIBUTING.md:222 points at it for the `contract_runner` convention and for keeping test docs in sync. Out of scope to write here, but it is a dangling reference.

**The suite needed tooling this container lacked:** `zstd`, `qmltestrunner` and the Qt6 QML modules were installed to run it, matching `tests.yml`. `patchelf` too, though it turns out not to matter — the fixture's fake `quickshell` is a shell script, so `libs_resolve_standalone` reports it self-sufficient and the patchelf branch is never taken. That branch remains uncovered; exercising it needs a real dynamically-linked binary with an unresolved `DT_NEEDED`, which I judged out of proportion here.

**Not verified on real hardware.** Everything above is fixture-driven against a fake `quickshell` and a stub `pacman`. No actual prebuilt release was downloaded, no real Arch host was used, and the network-fetch branch of `try_prebuilt` (the `curl` path when `WE_PREBUILT_DIR` is unset) is still untested — it is the production path, and it is the one thing this harness structurally cannot reach.


---

PR #106, opened 2026-08-04
fix(bar): bar window margins and standalone badge centring (#98, #99)

### LINKED ISSUE #98: Bar margins.bottom: operator precedence turns the dead-pixel workaround's -1 into +5
`modules/imi/bar/Bar.qml:149`

```qml
bottom: (Config.options.interactions.deadPixelWorkaround.enable && barRoot.anchors.bottom) * -1 || Config.options.bar.cornerStyle === 3 ? 5 : 0
```

`||` binds tighter than `?:`, so this parses as:

```qml
((A * -1) || (Config.options.bar.cornerStyle === 3)) ? 5 : 0
```

`A` is `deadPixelWorkaround.enable && barRoot.anchors.bottom`. When both hold, `A * -1` is `-1` — truthy — so the ternary yields **+5**, not the intended `-1`, for *every* corner style. The workaround's whole point is a 1px negative margin so the surface overhangs the edge Hyprland leaves a dead row on; instead the bar is pushed 5px *away* from the bottom edge, and Hyprland reserves 5px more with it (`exclusive + margin`, confirmed against `hyprctl monitors` on this machine: `reserved top = 45` for `exclusiveZone 40` + `margins.top 5`).

`margins.right` on the line above has the correct shape (`... * -1` with no `||`), which is what the bottom line was presumably meant to look like plus the M3 case.

### State that triggers it

`interactions.deadPixelWorkaround.enable: true` (default `false`, `Config.qml:1065`) **and** `bar.bottom: true`. Top bars are unaffected: `barRoot.anchors.bottom` is false, `false * -1` is `-0`, falsy, so the expression falls through to the `cornerStyle === 3` test correctly.

### Scope note

This predates today's margin work — it came in with `e2420f687` — but it is the one style-specific margin still living outside the two shared tokens that `131ce1165` introduced, and it sits directly in the exclusive-zone/margin path that commit reasoned about. Worth folding into the same cleanup.

**Severity: low.** Off by default, and only wrong for bottom bars, but silently defeats an opt-in workaround and adds 5px of reserved space when it fires.

### LINKED ISSUE #99: Standalone bar badges (timer, privacy, submap) are no longer concentric inside their group pill under Hug
`131ce1165` made `barPillHeight` asymmetric for Hug:

`modules/common/Appearance.qml:512-516`

```qml
property real barPillHeight: baseBarHeight - root.sizes.barMarginTop - root.sizes.barMarginBottom
property real barStandalonePillHeight: root.sizes.barPillHeight - root.sizes.barPillMargin
```

`barStandalonePillHeight` still subtracts one `barPillMargin` from the total, which only produces a concentric badge while the group pill's own inset is symmetric. Under Hug it no longer is.

Numbers, `baseBarHeight = 40`, `barPillMargin = 4`:

| style | group pill (BarGroup background) | standalone badge (`anchors.centerIn`) | inset |
|---|---|---|---|
| Hug (0) | `barMarginTop 0` / `barMarginBottom 4` → **y 0–36** | `barStandalonePillHeight` 32, centred in 40 → **y 4–36** | 4px top, **0px bottom** |
| Float / Islands / M3 | 4 / 4 → y 4–36 | 28, centred in 40 → y 6–34 | 2px / 2px |

So on a Hug bar the standalone badge sits flush against the bottom edge of the group pill it lives inside, instead of being inset evenly — the comment at `Appearance.qml:514-515` ("a touch shorter than the full group pill") describes the pre-change behaviour.

Three consumers, all `anchors.centerIn: parent` inside a full-bar-height root:

- `modules/imi/bar/TimerPill.qml:76` (pill at `:57-58`, `:76-77`)
- `modules/imi/bar/PrivacyIndicator.qml:92`
- `modules/imi/bar/SubmapIndicator.qml:64`

Before `131ce1165`, `barPillHeight` was `baseBarHeight - barPillMargin * 2` for every style, so all four were concentric. Hug regressed.

Reasoned from the geometry rather than observed — this machine runs `cornerStyle: 3`, where the badge is still concentric.

**Severity: low.** Cosmetic, 4px, and only on Hug — but Hug is the default `cornerStyle` (`Config.qml:901`), so it is the configuration most users see.

XephyLon:
## Describe your changes

Two bar-geometry bugs, both downstream of `131ce1165`'s margin work, both fixed by moving the arithmetic into `Appearance.sizes` where a unit test can evaluate it.

**#98 — `margins.bottom` operator precedence.** `(enable && anchors.bottom) * -1 || cornerStyle === 3 ? 5 : 0` parses as `((A * -1) || (cornerStyle === 3)) ? 5 : 0`, so a live dead-pixel workaround on a bottom bar yielded **+5** instead of −1 for every corner style — pushing the bar off the edge it was meant to overhang, and making Hyprland reserve 5px more with it. Split into `barDetachMargin` (M3's detachment, which is `hyprlandGapsOut` — that is what the `5` always was) and `barDeadPixelOverhang` (the 1px overhang), composed by `barBottomMargin`. They stay alternatives rather than addends, which is the priority the original `||` expressed.

**#99 — standalone badges no longer concentric under Hug.** `barStandalonePillHeight` subtracts one `barPillMargin` and the badge centres in the whole bar; that is only concentric while the group pill's own two insets match, and Hug's stopped matching. Adds `barStandalonePillOffset` (the group pill's centre relative to the bar's, mirrored by `bar.bottom`), applied by TimerPill / PrivacyIndicator / SubmapIndicator on **both** axes — BarGroup insets a horizontal bar's pill along its height and a vertical bar's along its width.

Two commits, one per issue.

Ready. Verification, and what I deliberately left alone, are in the review comment below.

Closes #98
Closes #99

### DIFF
--- a/dots/.config/quickshell/imi/modules/common/Appearance.qml
+++ b/dots/.config/quickshell/imi/modules/common/Appearance.qml
@@ -519,6 +519,36 @@ Singleton {
         // Standalone widget pills (Timer, Privacy) read as compact dynamic-island
         // badges: a touch shorter than the full group pill.
         property real barStandalonePillHeight: root.sizes.barPillHeight - root.sizes.barPillMargin
+        // The bar *surface's* own margins, as opposed to barMarginTop/Bottom
+        // above, which inset its body. Two terms, and they are alternatives
+        // rather than things to add up:
+        //
+        //   M3 (3) is the only style that detaches the whole surface from the
+        //   monitor edge, by the same gap Hyprland leaves around windows.
+        //
+        //   Hyprland leaves one untouchable pixel row on the right and bottom
+        //   screen edges. `interactions.deadPixelWorkaround` pulls the surface
+        //   1px *past* the edge so that row falls outside it - which replaces
+        //   the detach margin, since a surface already hanging over the edge
+        //   has no gap left to keep.
+        //
+        // Both used to be open-coded in Bar.qml as
+        //     (enable && anchors.bottom) * -1 || cornerStyle === 3 ? 5 : 0
+        // `||` binds tighter than `?:`, so the entire left-hand side was only
+        // the ternary's *condition*: with the workaround live that condition
+        // was truthy and the margin came out +5 for every style, pushing the
+        // bar away from the edge it was meant to overhang and making Hyprland
+        // reserve 5px more with it. They live here so the arithmetic is one
+        // expression a test can read rather than three inline ternaries.
+        property real barDetachMargin: Config?.options.bar.cornerStyle === 3
+            ? root.sizes.hyprlandGapsOut : 0
+        // One physical pixel, not a design token - it is the width of the row
+        // the compositor leaves out, so it does not scale with anything.
+        property real barDeadPixelOverhang: Config?.options.interactions.deadPixelWorkaround.enable
+            ? -1 : 0
+        property real barBottomMargin: (Config?.options.bar.bottom && root.sizes.barDeadPixelOverhang !== 0)
+            ? root.sizes.barDeadPixelOverhang
+            : root.sizes.barDetachMargin
         property real barCenterSideModuleWidth: Config.options?.bar.verbose ? 360 : 140
         property real barCenterSideModuleWidthShortened: 280
         property real barCenterSideModuleWidthHellaShortened: 190
--- a/dots/.config/quickshell/imi/modules/common/Appearance.qml
+++ b/dots/.config/quickshell/imi/modules/common/Appearance.qml
@@ -516,9 +516,23 @@ Singleton {
         // Follows from the two margins rather than assuming a symmetric inset,
         // so Hug's body reclaims the space its missing top margin frees.
         property real barPillHeight: baseBarHeight - root.sizes.barMarginTop - root.sizes.barMarginBottom
-        // Standalone widget pills (Timer, Privacy) read as compact dynamic-island
-        // badges: a touch shorter than the full group pill.
-        property real barStandalonePillHeight: root.sizes.barPillHeight - root.sizes.barPillMargin
+        // Standalone widget pills (Timer, Privacy, Submap) read as compact
+        // dynamic-island badges: a touch shorter than the group pill they sit
+        // inside, inset from it evenly on every side.
+        property real barStandalonePillMargin: root.spacing.space25
+        property real barStandalonePillHeight: root.sizes.barPillHeight
+            - root.sizes.barStandalonePillMargin * 2
+        // The badges centre themselves in the whole bar, but the group pill they
+        // sit inside is only centred there while its own two margins match -
+        // which stopped being true when Hug dropped its edge-side one, leaving
+        // the badge flush against the pill's window-side edge with all the slack
+        // on the other. This is the group pill's centre relative to the bar's,
+        // and it applies along whichever axis the bar's thickness runs:
+        // vertically for a horizontal bar, horizontally for a vertical one.
+        // `bar.bottom` puts the edge margin on the far side, so it mirrors.
+        property real barStandalonePillOffset: Config?.options.bar.bottom
+            ? (root.sizes.barMarginBottom - root.sizes.barMarginTop) / 2
+            : (root.sizes.barMarginTop - root.sizes.barMarginBottom) / 2
         // The bar *surface's* own margins, as opposed to barMarginTop/Bottom
         // above, which inset its body. Two terms, and they are alternatives
         // rather than things to add up:
--- a/dots/.config/quickshell/imi/modules/imi/bar/Bar.qml
+++ b/dots/.config/quickshell/imi/modules/imi/bar/Bar.qml
@@ -146,10 +146,14 @@ Scope {
                     right: true
                 }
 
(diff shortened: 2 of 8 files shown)


XephyLon:
## What changed

**`4242c02` — #98, the `margins.bottom` precedence bug**

`Bar.qml`'s three window margins are now bare token references:

```qml
margins {
    top: Appearance.sizes.barDetachMargin
    right: Appearance.sizes.barDeadPixelOverhang
    bottom: Appearance.sizes.barBottomMargin
}
```

with the arithmetic in `Appearance.sizes`:

| token | value |
|---|---|
| `barDetachMargin` | `cornerStyle === 3 ? hyprlandGapsOut : 0` |
| `barDeadPixelOverhang` | `deadPixelWorkaround.enable ? -1 : 0` |
| `barBottomMargin` | `bar.bottom && overhang !== 0 ? overhang : detach` |

Two judgement calls worth flagging:

- **The `5` became `hyprlandGapsOut`, not a spacing token.** It is not on the M3 scale (…4, 6…) and it is not UI rhythm — it is the gap Hyprland leaves around windows, which is why `BarContent.qml` already writes M3's inset as `Config.options.hyprland.general.gapsOut || 5`. Value is unchanged at 5. Naming it `space75` (6) would have silently moved the bar 1px.
- **The overhang keeps beating the detach margin**, rather than the two adding. That is what the original `||` did, and it is the behaviour the issue reasons about ("`margins.right` … is what the bottom line was presumably meant to look like plus the M3 case"). It means M3 + workaround + bottom bar loses M3's 5px detachment — arguably correct, since a surface hanging over the edge has no gap left to keep, but it is a behaviour choice and not forced by the bug.
- **`-1` stayed a literal.** I first reached for `borderWidth.standard` and backed it out: it is one physical pixel of compositor dead row, not a border, and AGENT.md's own `colLayer0`/`colLayer1` note is about exactly that kind of by-analogy token pick.

**`699e91a` — #99, badge concentricity**

`barStandalonePillMargin` (`space25`) names the inset, `barStandalonePillHeight` derives from it (numerically identical to before — the height was never the bug), and the new `barStandalonePillOffset` shifts each badge onto its group pill's centre. Applied on both axes in all three badges, and mirrored by `bar.bottom`.

## How I verified it

**Every test added was proven by mutation** — written, run against the *broken* code, seen to fail, then reverted. Neither pin is one I only watched pass.

*#98* — transplanted the original expression verbatim into `barBottomMargin`:

```
FAIL!  : BarGeometryTest::test_deadPixelOverhangBeatsTheM3DetachMargin()
         the workaround must pull the surface past the edge, cornerStyle 0
   Actual   (): 5
   Expected (): -1
```

That is the issue's exact claim (+5 where −1 belongs) reproduced by the test as a first-class failure, not inferred.

*#99* — set `barStandalonePillOffset: 0`, i.e. the pre-fix "centre in the whole bar":

```
FAIL!  : BarGeometryTest::test_standaloneBadgeIsConcentricInsideItsGroupPill()
         badge is not concentric in its group pill (cornerStyle 0, bottom false)
   Actual   (): 4
   Expected (): 0
FAIL!  : BarGeometryTest::test_hugIsTheStyleThatNeedsTheOffset()
```

4px above / 0 below, matching the issue's table row for Hug exactly. I also mutated the *height* back to `barPillHeight - barPillMargin` on its own and the suite stayed green — correctly, since that expression is numerically identical. Worth stating because it is the assertion that could have looked load-bearing and isn't.

*Contract pins* — restored the inline `margins` block and separately deleted TimerPill's offsets; `test_bar_geometry_contract.py` failed on each, naming the offending expression.

**Suites:** `260 passed, 0 failed` before any change → `265 / 0` after commit 1 → `268 / 0` after commit 2. Green before each commit, never fewer than baseline.

- **No live-shell verification.** This environment has no Hyprland, no Wayland display and no `qs`, so CONTRIBUTING's log-grep loop and the `DesignSystemCompile` check (which skips without `WAYLAND_DISPLAY`) never ran. `qmllint` reports no syntax errors and `Appearance.qml` is instantiated for real by the QML suite, but **nobody has looked at a bar**. The 4px is reasoned from the same geometry the issue reasons from. Worth an eyeball on a Hug bar with a timer running before merge.
- **The vertical-bar half of #99 is reasoned, not observed.** Derived from `BarGroup.qml`'s own `leftMargin`/`rightMargin` mapping. I included it because it is the same defect from the same commit and the fix is the same token — but the issue only worked the horizontal numbers, and a badge's width in the vertical bar is content-derived, so the symptom there is subtler than a flush edge.
- **Did not touch `margins.right`.** It is correct as written and #98 explicitly cites it as the reference shape; it moved to a token only so all three margins read alike.
- **Did not give `margins.top` a dead-pixel term.** The dead row is bottom/right only (`Config.qml:1064`), and the original code did not have one either.
- **Did not widen the M3 `5` to read `Config.options.hyprland.general.gapsOut`** the way `BarContent.qml` does. That would change behaviour for anyone with a non-default `gapsOut`, which is a separate decision from this bug.
- **`superpowers` was not available.** CONTRIBUTING.md makes it a hard requirement; it is not in this session's skill listing and there is no marketplace to install from in this container. Flagging rather than silently proceeding, as that section asks.

---

PR #107, opened 2026-08-04
fix(sddm): restore the whole matugen block, and only uninstall a theme we installed

### LINKED ISSUE #100: Uninstall step 5 fires on a pre-fork ii-sddm-theme install ImI never made, and deletes it
## What breaks

Uninstall step 5 decides to run the SDDM theme's uninstaller based only on whether a theme *directory* exists — including the pre-fork name, which ImI may never have installed.

`sdata/subcmd-uninstall/0.run.sh:23`:

```sh
if [[ -d /usr/share/sddm/themes/imi-sddm-theme || -d /usr/share/sddm/themes/ii-sddm-theme ]]; then
```

`ii-sddm-theme` is upstream [3d3f/ii-sddm-theme](https://github.com/3d3f/ii-sddm-theme)'s install name. A user who installed upstream's theme on their own — years before ImI, by its own `setup.sh` — has that directory. If they then install ImI *without* the SDDM extra (`INSTALL_SDDM=0`, the default) and later run `setup uninstall`, this branch fires and hands their machine to our uninstaller, which removes things ImI never created:

- `/usr/share/sddm/themes/ii-sddm-theme` (`uninstall.sh:103-107`)
- `/etc/sddm.conf.d/ii-sddm-theme.conf` (`:114-117`) — the file carrying their `Current=`
- `~/.config/ii-sddm-theme` (`:144-147`)
- `/usr/share/fonts/ii-sddm-theme-fonts` (`:152-162`)
- their `[templates.iisddmtheme]` matugen block (`:166-190`)
- `/etc/sudoers.d/sddm-theme-<user>` (`:194-203`)

Losing the theme directory while `Current=` still names it is the login-blocking case described in XephyLon/imi-sddm-theme#2.

## The repo already knows the right pattern

Immediately below, step 4 does exactly the right thing (`0.run.sh:41-44`):

```sh
# Step 4: ... Only remove it if it is actually ours (carries the
# immaterial-impulse marker), never a foreign /usr/local/bin/quickshell.
if [[ -f /usr/local/bin/quickshell ]] && grep -q "immaterial-impulse" /usr/local/bin/quickshell 2>/dev/null; then
```

Step 5 has no equivalent ownership check.

## Severity

**Medium**, escalating to login-blocking on the affected machine (their `Current=` survives in `kde_settings.conf` / `/etc/sddm.conf` while the theme directory is gone). Low frequency — needs a pre-existing upstream install plus an ImI uninstall — but the blast radius is somebody else's working login screen, and the only confirmation they get is the theme's own generic `This script will remove imi-sddm-theme and all its components. Continue? [y/n]`, which does not describe what is actually about to be deleted.

## Fix direction

Gate step 5 on evidence that *we* installed it — e.g. the `INSTALLED_LISTFILE` entry, or a marker written by step 5, or at minimum only fire on `imi-sddm-theme` (our name) and mention the legacy name in the "remove manually" hint instead of acting on it.

## Adjacent note (not filed separately)

The theme's `uninstall.sh` has no equivalent of the `IMI_SDDM_ASSUME_YES` that `99b2e38` added to `setup.sh` — its confirmation is a bare `read` (`uninstall.sh:81`). On a tty (the normal path, since `0.run.sh:13` calls `pause` first) this is fine; driven with stdin closed it would EOF, hit the `*)` case, print `Uninstallation aborted by user` and `exit 0`, and the hub's `|| printf ...` would not notice because the exit status is 0.

### LINKED ISSUE #101: restore_sddm_matugen_hook restores the SDDM post_hook but not the [templates.iisddmtheme] block, so greeter colours stop regenerating after every update
## What breaks

`restore_sddm_matugen_hook` (`sdata/subcmd-install/3.files-legacy.sh:17-53`) exists because the dots sync overwrites `~/.config/matugen/config.toml` and wipes what the SDDM theme's installer appended to it. It restores the **trigger** but not the **template**, so after any `setup install` the greeter's colours stop being regenerated.

The theme's installer writes one block (`imi-sddm-theme` `setup.sh:573-579`):

```toml
[templates.iisddmtheme]
input_path = '~/.config/imi-sddm-theme/SddmColors.qml'
output_path = '~/.config/imi-sddm-theme/Colors.qml'
post_hook = 'python3 ~/.config/imi-sddm-theme/generate_settings.py && sudo ~/.config/imi-sddm-theme/sddm-theme-apply.sh &'
```

The dots sync (`3.files-legacy.sh:59-65` → `install_dir__sync "dots/.config/matugen" ...`) replaces `config.toml` with `dots/.config/matugen/config.toml`, deleting the whole block. `restore_sddm_matugen_hook` then puts back only the `post_hook`, and under `[config]` rather than under the template (`3.files-legacy.sh:36-51`). The `input_path`/`output_path` pair is never restored.

Result: `~/.config/imi-sddm-theme/Colors.qml` is no longer generated from `SddmColors.qml` on a wallpaper change. The hook still fires, but there is nothing regenerating the colours it is supposed to publish — the greeter's palette freezes at whatever it was when the theme was last installed.

## This is the live state on the machine I reviewed on

```
$ grep -n "post_hook\|^\[" ~/.config/matugen/config.toml
1:[config]
2:post_hook = 'python3 ~/.config/imi-sddm-theme/generate_settings.py && sudo ~/.config/imi-sddm-theme/sddm-theme-apply.sh &'
5:[templates.m3colors]
...
```

No `[templates.iisddmtheme]`. `~/.config/imi-sddm-theme/SddmColors.qml` is still on disk, unused.

## Second-order

- The guard at `3.files-legacy.sh:34` is `grep -q '^post_hook' "$matugen_conf" && return 0`. The theme's own block also has `post_hook` at column 0, so on a config where the theme block *survived*, restore correctly no-ops — but that same pattern means the function cannot distinguish "template block intact" from "only my restored `[config]` hook present", so it never notices the half-restored state it created.
- A `[config] post_hook` is global: it now runs after *every* matugen invocation, not just the SDDM template's, so the apply script (and a `sudo` call) is triggered by every theme regeneration.
- The restored hook targets `~/.config/imi-sddm-theme/sddm-theme-apply.sh`, which currently fails immediately for an unrelated reason — XephyLon/imi-sddm-theme#1.

## Severity

**Medium.** Silent and persistent: every ImI update leaves the greeter's colours stuck, with a restore function that reports success (`restored the SDDM theme's matugen post_hook`). Recovery requires re-running the theme's installer, which nothing tells the user to do.

## Fix direction

Restore the full `[templates.iisddmtheme]` block (input/output/post_hook) rather than a bare `[config] post_hook`, keyed off which theme directory is present exactly as the current mode detection already does — or exclude `matugen/config.toml` from the `--delete` sync and merge instead.

## Scope note

Found while reviewing the SDDM install/uninstall chain; the file itself is `3.files-legacy.sh` rather than `5.sddm-theme.sh`, but the defect is entirely about the SDDM theme's integration.

XephyLon:
Closes #101. Closes #100.

Two halves of the same subject: what the dots sync does to the SDDM theme's matugen integration, and what the uninstaller does to an SDDM theme that may not be ours.

## #101 — the matugen block was restored in pieces

`restore_sddm_matugen_hook` exists because install step 3 deploys `dots/.config/matugen/config.toml` with `rsync --delete`, wiping the `[templates.iisddmtheme]` block the theme's installer appends there. It put back only the `post_hook`, and under `[config]` rather than under the template.

That is worse than restoring nothing: the hook still fires — now after *every* matugen invocation, dragging a `sudo` call with it — while the `input_path`/`output_path` pair that regenerates `Colors.qml` is gone. Nothing produces the file the hook is meant to publish, so the greeter's palette freezes on every ImI update while the function reports success. Its guard (`grep ^post_hook`) matched its own half-restore, so the bad state was permanent once created.

Now: the guard is the block header; the full block is written in the same shape `setup.sh` writes; and a stray global `post_hook` naming the theme's apply script is deleted first (that line is on every machine this ran on). A `post_hook` that doesn't name the apply script is the user's and is left alone.

Two further defects surfaced while fixing it, both from the theme's recent history:

- **The detection marker no longer existed.** It keyed off `~/.config/<theme>/sddm-theme-apply.sh`, and that file is not installed there any more — the apply script is now root-owned at `/usr/local/lib/<theme>/`, because a NOPASSWD sudoers rule names it and sudo matches by path. On any current install the marker matched nothing and the restore silently never ran. It now keys off `SddmColors.qml`, the template's own `input_path`.
- **The restored hook named a path sudo would not accept.** Same reason. A password prompt from a hook backgrounded with `&` is a hang, not an error. The privileged path is preferred, with the in-config one kept as a fallback for installs predating the move, whose sudoers rule still names it.

## #100 — uninstall step 5 fired on somebody else's theme

It gated on a directory existing, under either name. `ii-sddm-theme` is upstream `3d3f/ii-sddm-theme`'s install name, so a user who installed upstream's theme themselves, then installed ImI *without* the SDDM extra (`INSTALL_SDDM=0`, the default), had their theme directory, drop-in, config dir, fonts, matugen block and sudoers rule deleted by an ImI uninstall. Losing the directory while a `Current=` still names it is the login-blocking case.

Step 4 immediately below already models the right shape — prove ownership, don't infer it. Step 5 now does the same:

- install step 5 writes a marker on a successful hand-off, under `XDG_STATE_HOME` so the dots sync (which `rsync --delete`s `~/.config`) can't erase the record. Written only in the success branch.
- uninstall fires on the marker, or on `/usr/share/sddm/themes/imi-sddm-theme` — our name, only this fork installs under it. Installs predating the marker still uninstall cleanly.
- the marker alone authorises the pre-fork name (we installed it before the rename and the migration hasn't run).
- a bare pre-fork directory with no marker is **reported, never acted on**, including the warning that deleting it while a `Current=` names it costs the user their login screen.

The marker is removed only when the theme's uninstaller succeeds — newly meaningful, since that uninstaller used to exit 0 when it removed nothing (fixed in the satellite PR). The manual-removal hint also still named pre-rename paths and omitted `/usr/local/lib/imi-sddm-theme`.

## Tests

`260 passed, 0 failed`.

- `test_sddm_matugen_hook_restore.py` — extended. Its existing cases pinned the half-restore as correct (they asserted the in-home hook path and never checked that `input_path`/`output_path` came back), so those assertions are replaced rather than kept; that was the test being wrong, not the change being a regression. 13 of the 17 cases fail against the previous implementation.
- `test_sddm_uninstall_ownership.py` — new. Drives the ownership gate for every combination of marker / current name / pre-fork name, and pins that install and uninstall agree on the marker path. Fails loudly against the previous implementation.

## Not done here

`SDDM_REF` and `WE_REF` are untouched. The pin needs to move for the satellite fixes to reach anyone — the SHA is in the comment below.

### DIFF
--- a/dots/.config/quickshell/imi/tests/run_tests.sh
+++ b/dots/.config/quickshell/imi/tests/run_tests.sh
@@ -288,9 +288,10 @@ if ! python3 "$SCRIPT_DIR/test_kboptions_migration_runtime.py"; then
     exit 1
 fi
 
-# The hook is a path into the theme's directory, and that directory was
-# renamed - a wrong path here means the login screen silently stops following
-# the wallpaper.
+# The block is a set of paths into the theme's directory and at the apply
+# script sudo will accept; the directory was renamed and the script moved. A
+# wrong path here means the login screen silently stops following the
+# wallpaper, and restoring only part of the block is how it stopped last time.
 echo "Running SDDM matugen hook restore tests..."
 if ! python3 "$SCRIPT_DIR/test_sddm_matugen_hook_restore.py"; then
     echo "SDDM matugen hook restore tests failed."
--- a/dots/.config/quickshell/imi/tests/run_tests.sh
+++ b/dots/.config/quickshell/imi/tests/run_tests.sh
@@ -298,6 +298,15 @@ if ! python3 "$SCRIPT_DIR/test_sddm_matugen_hook_restore.py"; then
     exit 1
 fi
 
+# Uninstall step 5 hands the machine to the theme's uninstaller, which removes
+# the drop-in carrying Current= along with the theme. Firing it on a theme we
+# did not install is somebody else's login screen.
+echo "Running SDDM uninstall ownership tests..."
+if ! python3 "$SCRIPT_DIR/test_sddm_uninstall_ownership.py"; then
+    echo "SDDM uninstall ownership tests failed."
+    exit 1
+fi
+
 echo "Running SDDM theme source tests..."
 if ! python3 "$SCRIPT_DIR/test_sddm_theme_source.py"; then
     echo "SDDM theme source tests failed."
--- a/dots/.config/quickshell/imi/tests/test_sddm_matugen_hook_restore.py
+++ b/dots/.config/quickshell/imi/tests/test_sddm_matugen_hook_restore.py
@@ -1,22 +1,38 @@
 #!/usr/bin/env python3
-"""`restore_sddm_matugen_hook` must point at the theme that is actually there.
+"""`restore_sddm_matugen_hook` must restore the whole block, pointing at the
+theme that is actually there and at the apply script sudo will actually accept.
 
 We ship our own `~/.config/matugen/config.toml` and deploy it with
-`rsync --delete`, which removes the SDDM theme's `post_hook` on every update -
-so the login screen stops following the wallpaper until the hook is put back.
-`sdata/subcmd-install/3.files-legacy.sh` restores it.
+`rsync --delete`, which removes the SDDM theme's `[templates.iisddmtheme]` block
+on every update - so the login screen stops following the wallpaper until it is
+put back. `sdata/subcmd-install/3.files-legacy.sh` restores it.
 
-The hook is a path, and the theme's directory name changed: it installs as
-`imi-sddm-theme` now, and installs from before that rename are still under
-`ii-sddm-theme` until the theme's own installer migrates them. So the restore
-has to resolve the name rather than hardcode it, and it has to prefer the new
-one - during a migrating update both directories exist for a moment, and a hook
-written against the one about to be deleted is broken the instant the migration
-finishes.
+Three things have to be right, and each has been wrong:
+
+1. **The whole block, not just the hook.** The restore used to write a bare
+   `post_hook` under `[config]`, which is worse than writing nothing: the hook
+   fires - now after every matugen run, not just this template's - while the
+   `input_path`/`output_path` pair that regenerates `Colors.qml` is gone, so
+   there is nothing for it to publish. The greeter's palette froze and the
+   function reported success (issue #101). Its guard (`grep ^post_hook`) then
+   matched its own half-restore forever, so it never corrected the state it had
+   created.
+
+2. **The theme's directory name.** It installs as `imi-sddm-theme` now, and
+   installs from before that rename are still under `ii-sddm-theme` until the
+   theme's own installer migrates them. During a migrating update both exist for
+   a moment, and a block written against the one about to be deleted is broken
+   the instant the migration finishes.
+
+3. **The apply script's path.** It is installed root-owned at
+   `/usr/local/lib/<theme>/`, because the NOPASSWD sudoers rule names it and
+   sudo matches by path - a rule pointing anywhere the user can write is
+   equivalent to `NOPASSWD: ALL`. A hook naming any other location prompts for a
+   password from a backgrounded hook, which is a hang rather than an error.
 
 Hermetic: sources just this function out of 3.files-legacy.sh (the file is
 meant to be sourced by ./setup and has a procedural body) and runs it against a
-sandbox HOME.
+sandbox HOME, with the privileged directory rewritten into that sandbox.
 """
 
 import subprocess
@@ -35,6 +51,13 @@ SCRIPT = ROOT / "sdata/subcmd-install/3.files-legacy.sh"
 FUNCTION = "restore_sddm_matugen_hook"
 
 BASE_CONF = "[config]\nreload_apps = true\n"
+BLOCK_HEADER = "[templates.iisddmtheme]"
+
+# The apply script's real home. The harness rewrites this to reach the sandbox;
+# test_the_privileged_path_is_where_we_think_it_is keeps the rewrite honest, so
+# a change to the install path fails loudly instead of silently exercising the
+# real /usr/local/lib.
+PRIV_LITERAL = "/usr/local/lib/${theme_name}"
 
 
 def _extract_function(text, name):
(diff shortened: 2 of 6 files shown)


XephyLon:
## What changed

Two commits, on `main` at `a0a985b1` (0.14.9).

**`3064338` — `fix(install): restore the whole SDDM matugen block, not just its post_hook`** (#101)

`sdata/subcmd-install/3.files-legacy.sh`. The restore now writes the full `[templates.iisddmtheme]` block — `input_path`, `output_path`, `post_hook` — in the same shape the theme's `setup.sh:630-636` writes, so a later re-run of the theme installer replaces it rather than appending a second one. Three separate defects, only the first of which is what #101 describes:

1. **Half a block.** A bare `[config] post_hook` with no `input_path`/`output_path` fires forever and regenerates nothing. And the guard was `grep '^post_hook'`, which its own half-restore satisfies — so once the state existed the function could never notice or correct it. The guard is now `^\[templates\.iisddmtheme\]`, and a stray global `post_hook` naming the theme's apply script is deleted before the block is written. That line is on every machine this ever ran on. A `post_hook` that doesn't name the apply script is the user's and is untouched.

2. **The detection marker no longer exists on any current install.** It keyed off `~/.config/<theme>/sddm-theme-apply.sh`. That file is not installed there any more — `setup.sh:413-418` installs the apply script root-owned at `/usr/local/lib/<theme>/` and `rm -f`s the staged copy, because the NOPASSWD sudoers rule names it and sudo matches by path. So on a current install the marker matched nothing and **the restore silently never ran at all**. It now keys off `SddmColors.qml`, which is the template's own `input_path` — the file whose absence would make the block pointless.

3. **The restored hook named a path sudo would reject.** Same root cause. A password prompt from a hook backgrounded with `&` is a hang, not an error message. The privileged path is preferred; the in-config path is kept as a fallback for installs made before the move, whose sudoers rule still names it. This also means the fix is correct on both sides of wherever `SDDM_REF` happens to sit.

**`c9aa44b` — `fix(uninstall): only remove an SDDM theme we actually installed`** (#100)

`sdata/subcmd-uninstall/0.run.sh` + `sdata/subcmd-install/5.sddm-theme.sh`. Step 5 gated on a directory existing under either name; `ii-sddm-theme` is upstream's install name. Ownership is now proved the way step 4 already does it:

- step 5 writes `${XDG_STATE_HOME:-$HOME/.local/state}/immaterial-impulse/sddm-theme-installed` on a successful hand-off (`ref=` + `installed=`). State, not config — the dots sync `rsync --delete`s `~/.config`, so a marker there would erase itself. Written only inside the success branch.
- uninstall fires on the marker, or on `/usr/share/sddm/themes/imi-sddm-theme` (our name; only this fork installs under it, so installs predating the marker still work).
- the marker alone authorises the pre-fork name — we installed it before the theme renamed itself and its migration hasn't run.
- **a bare pre-fork directory with no marker is reported, never acted on**, naming the directory, saying it's most likely upstream's, and warning that deleting it while a `Current=` names it leaves no graphical login.

The marker is removed only when the theme's uninstaller *succeeds*. That branch is newly meaningful: the theme's `uninstall.sh` used to exit 0 when it removed nothing, which is the adjacent note in #100 and is fixed in [imi-sddm-theme#8](https://github.com/XephyLon/imi-sddm-theme/pull/8). The fallback hint also still named pre-rename paths and omitted `/usr/local/lib/imi-sddm-theme` — a hint that leaves a NOPASSWD-adjacent directory behind isn't a complete hint.

## How this was verified — without running any destructive script

Nothing here was run against a real system. No `setup`, no `setup uninstall`, no theme `setup.sh`/`uninstall.sh`/`check.sh`.

**Full suite green:** `cd dots/.config/quickshell/imi/tests && bash run_tests.sh` → **`Totals: 260 passed, 0 failed, 0 skipped`**, `All tests passed successfully!`. Run twice — once before committing, once after — with `QT_QPA_PLATFORM=offscreen` and the Qt6 packages CI installs.

**The tests themselves are the verification**, and both were confirmed to fail against the previous code rather than passing vacuously (CONTRIBUTING's "prove a new static check can fail"):

- `test_sddm_matugen_hook_restore.py` — extracts `restore_sddm_matugen_hook` out of the sourced-only install script, rewrites the `/usr/local/lib/${theme_name}` literal into a tempdir, and drives it against fabricated installs: full block restored, hook under the template rather than `[config]`, stray global hook cleaned up, a foreign `post_hook` left alone, the privileged vs. in-home apply path, matugen-only vs. ii-matugen mode, name resolution and preference, idempotency, and the three cases where it must write nothing. With `3.files-legacy.sh` stashed back to its previous version: **13 of 17 fail.**
- `test_sddm_uninstall_ownership.py` (new) — extracts the ownership gate, rewrites `/usr/share/sddm/themes` into a tempdir, and runs every combination of marker × current name × pre-fork name, asserting the decision and the user-facing text. Plus static pins that install and uninstall agree on the marker path, that the marker is written only in the success branch, that it's removed only on a zero exit, and that the fetch-and-run is actually inside the gate. With both scripts stashed back: fails at import with `could not extract the step-5 ownership gate`, i.e. loudly, not silently.

Both are registered in `run_tests.sh` (`python3 <file>` with a non-zero-exits guard), and both subclass `unittest.TestCase` with `unittest.main()` — not the bare-`test_*`-functions shape that has shipped as a silent no-op here three times.

**Static cross-checks** against the satellite at `cd56fcd`: the emitted block's key names, order and quoting match `setup.sh:630-636`; the privileged apply path matches `setup.sh:50-51`; `SddmColors.qml` is confirmed as `MATUGEN_QML_INPUT_TEMPLATE` (`setup.sh:57`) and is genuinely left in the config dir.

**`/etc` confirmation:** nothing in this repo's changes touches `/etc`, and nothing ran that could. While reviewing the satellite side I did source its `check.sh` into a sandbox — with a `sudo` stub first on a `env -i` `PATH`. This container has **no `/etc/sddm.conf` and no `/etc/sddm.conf.d`**, verified before and after; neither was created. Details in the [satellite PR comment](https://github.com/XephyLon/imi-sddm-theme/pull/8).

- **Did not touch `SDDM_REF` or `WE_REF`.** Both untouched in both hub files. See below.
- **Did not touch `AGENT.md`** — it has no SDDM section, and neither change alters a mechanism it describes. `docs/tests-README.md` doesn't enumerate these checks either.
- **Did not add an `IMI_SDDM_ASSUME_YES` path to the hub's uninstall.** Step 5 keeps the interactive confirmation; the satellite change makes a declined run *reportable*, not skippable.
- **Did not change what step 5 removes**, only whether it fires. The hand-off to the theme's uninstaller is unchanged.
- **Did not backfill a marker for existing installs.** Unnecessary — `/usr/share/sddm/themes/imi-sddm-theme` already authorises removal on its own; the marker only adds the pre-fork-name case.
- **Did not exclude `matugen/config.toml` from the `--delete` sync**, the alternative #101 offers. Restoring the block keeps one shipped `config.toml` as the source of truth for our own templates; a merge-instead-of-sync would leave stale ImI template blocks behind forever on any user who once had them.
- **Did not "fix" the `[templates.iisddmtheme]` key name.** It's a matugen template id, not a path; renaming it orphans the block on every existing install and desyncs from the theme's `setup.sh`/`uninstall.sh`.

## The SHA the hub's `SDDM_REF` should move to

Currently `cd56fcdcd70a0800918262e6cba2ebcc5aa9b447` (imi-sddm-theme 0.1.1) in both files. It should move to the commit that lands on `imi-sddm-theme`'s `main` from [imi-sddm-theme#8](https://github.com/XephyLon/imi-sddm-theme/pull/8) — branch head today:

```
27c93f73450b9b8e679a6ea0517eb6a335e40c40
```

**If that PR is rebase-merged the resulting `main` commit will have a different SHA** — use `git rev-parse origin/main` in imi-sddm-theme after merging, not the SHA above.

Set the same value in **both**:

- `sdata/subcmd-install/5.sddm-theme.sh` → `SDDM_REF="${SDDM_REF:-<sha>}"`
- `sdata/subcmd-uninstall/0.run.sh` → `_sddm_ref="${SDDM_REF:-<sha>}"`

Full 40-char SHA, never a tag (`test_the_pin_is_a_full_commit_sha`, `test_the_uninstaller_is_pinned_to_the_same_commit`).

**Order matters.** `c9aa44b` makes uninstall step 5 branch on the theme uninstaller's exit status and remove the marker only on success. That status is only meaningful from imi-sddm-theme `27c93f7` onward. Merging this PR while the pin still points at `cd56fcd` means the hub calls an uninstaller that exits 0 on a decline, so a declined uninstall would clear the marker for a theme still on disk — recoverable (the `imi-sddm-theme` directory name still authorises removal by itself), but it defeats the point. **Move the pin in the same series as merging this.**


---

Issue #94, opened 2026-08-03
Background: a failed WE project keeps the transition overlay up, hiding the static fallback for ~9s


XephyLon:
## What breaks

991e099aa added a static-image fallback for a Wallpaper Engine project the renderer gives up on (`weFailed` → `weShown` false → the plain `wallpaper` image becomes visible). The fallback chain itself is correct — I traced `wallpaper`, `lockPeel`, the centred-wallpaper loader and `PluginWidget.weSurfaceItem`, and all of them fall back the right way.

But the WE **transition** overlay is not part of that chain, and it is drawn above the fallback:

`dots/.config/quickshell/imi/modules/imi/background/Background.qml:557-570`

```qml
ShaderEffect {
    id: weTransition
    z: 1
    visible: bgRoot.weTransitioning
    property var fromImage: weOldStill
    property var toImage: weLiveSource
```

`z: 1` puts it above `wallpaper` (:601, default z). Its only visibility term is `weTransitioning` — no `weShown`, no `weFailed`, no null check on `weLoader.item`. So while a transition is armed, the frozen snapshot of the *previous* project covers the static fallback that `weFailed` just revealed.

## Why the failure path lands exactly here

The way `failed` latches is by loading a project that cannot render — which is precisely when a transition has just been armed. In `qs-wallpaperengine` v0.2.0, `setProjectPath` clears `rendered` and clears `failed`, and the failure is reported from `updatePaintNode` on the early-return path *before* a texture is ever acquired. So for a failing project, `rendered` goes true→false and never comes back.

Trace, with `weLoadedProject === A` and the user switching to a project `B` that fails:

1. `loadWeWallpaper(B)` takes the transition branch (:284-294): snapshots A into `weOldStill`, sets `weTransitionProgress = 0.0`, `weTransitioning = true`, assigns `projectPath = B`.
2. `renderedChanged` fires for true→false. The handler at :525-534 requires `weLoader.item.rendered` truthy, so it does nothing. `weTransitionDelay` and `weTransitionAnim` never start.
3. `failedChanged` fires. `weFailed` (:204) → true, `weShown` (:208) → false, `wallpaper.visible` (:601) → true. The static fallback is drawn — and immediately covered.
4. `weTransitionProgress` stays `0.0`, so `weTransition` paints project A's last frame full-screen over it.
[...]

Issue #92, opened 2026-08-03
Background: wePendingProject is never invalidated, so un-suppressing can replay a superseded Wallpaper Engine project


XephyLon:
## What breaks

`wePendingProject` is the stash for a Wallpaper Engine project switch that arrives while the background's contents are suppressed (fullscreen + `hideWhenFullscreen`). It is written in exactly one place and cleared in exactly one place — when it is replayed. Nothing invalidates it when the request it holds is superseded or when the WE layer it was destined for goes away, so un-suppressing can load a project the user already moved off.

The hole is the early return on the first line of `loadWeWallpaper`:

`dots/.config/quickshell/imi/modules/imi/background/Background.qml:262`

```qml
if (!weLoader.item || path === bgRoot.weLoadedProject) return
```

Both arms return **before** the suppression check at :267, so neither one gets the chance to clear a stash that is now wrong.

## Concrete state that triggers it

Preconditions: a true-fullscreen window on the monitor's active workspace, `background.hideWhenFullscreen` true (the default, `defaults/config.json:258`), screen not locked. After the 400 ms `fullscreenSuppressDelay` (:128-132) `suppressContents` is true. A WE project `A` is loaded, so `weLoadedProject === "A"`.

1. Switch to project `B`. `onWeProjectPathChanged` (:296) → `loadWeWallpaper("B")` → suppressed → `wePendingProject = "B"` (:268).
2. Switch back to `A`. `loadWeWallpaper("A")` → `path === bgRoot.weLoadedProject` → **returns at :262**. `wePendingProject` is still `"B"`.
3. Leave fullscreen. `suppressedForFullscreen` clears, `suppressContents` goes false, `onSuppressContentsChanged` (:159-168) replays `"B"`.

The desktop now renders `B` while `Config.options.wallpaperSelector.wallpaperEngine.activePath` and `weProjectPath` both say `A`. Nothing reconciles them; it persists until the user makes another *distinct* switch.

Second route to the same end, no undo needed: with a stash pending, anything that drops `weActive` (:193 — work-safety trigger, selecting a static wallpaper, selecting a `web`-type project) deactivates `weLoader`. On reactivation `onLoaded` (:516-519) re-syncs `weLoadedProject` to `weProjectPath`, but `wePendingProject` is untouched and still holds the stale path, so the next un-suppress replays it.

## Also worth folding in

[...]

PR #108, opened 2026-08-04
fix(background): Wallpaper Engine transition and pending-project state (#94, #92)

XephyLon:
## Describe your changes

Two defects in `Background.qml`'s Wallpaper Engine state, both in the same area and taken together because they interact.

**#92 — `wePendingProject` was never invalidated.** Written in one place, cleared in one place (when replayed), and both arms of `loadWeWallpaper`'s first line return *before* the suppression check, so neither could clear a stash that had just been superseded. Un-suppressing could load a project the user had already moved off. The stash is removed rather than given invalidation: what un-suppressing wants to apply is always the *current* request, `weProjectPath`, which cannot go stale. That also disposes of the `""`-as-both-sentinel-and-legal-path overload the issue flags.

**#94 — the transition overlay outlived its own destination.** `weTransition` is `z: 1` above the static wallpaper and its only visibility term was `weTransitioning`. A project the renderer gives up on latches `failed` exactly when a transition has just been armed, so the peel painted a frozen still of the *outgoing* project over the static fallback that `weFailed` had just revealed — for the watchdog's full ~9.2s. Fixed on both sides: `visible` gains `!weFailed` and a null check, and `onWeFailedChanged` / `onWeActiveChanged` settle the transition immediately through a shared `settleWeTransition()` that the watchdog now also uses.

Two commits, one per issue.

Ready, with one thing worth your call — see the review comment below: #92's fix deletes state rather than adding invalidation, which is a slightly larger change than the issue proposes.

Refs #92
Refs #94

### DIFF
--- a/dots/.config/quickshell/imi/modules/imi/background/Background.qml
+++ b/dots/.config/quickshell/imi/modules/imi/background/Background.qml
@@ -270,6 +270,32 @@ Variants {
         property real weTransitionProgress: 1.0  // 0 = old still, 1 = new surface
         property bool weTransitioning: false
 
+        // Put the transition back to its resting state: no peel on screen, the
+        // snapshot released, progress at "new surface". Reached from the
+        // watchdog below, and from the two states that make the first frame the
+        // peel is waiting for impossible to ever get.
+        function settleWeTransition() {
+            weTransitionAnim.stop()
+            weTransitionDelay.stop()
+            bgRoot.weTransitionProgress = 1.0
+            bgRoot.weTransitioning = false
+            weOldStill.source = ""
+        }
+        // `failed` latches on a project the renderer cannot draw, and that is
+        // precisely a moment when a transition has just been armed: the switch
+        // snapshots the outgoing frame and then waits on a frame that will never
+        // arrive, since the failure is reported before a texture is ever
+        // acquired. Nothing in the transition's own state can see that, so the
+        // peel used to sit there for the watchdog's whole budget - about nine
+        // seconds of the wallpaper the user just switched *away from*, painted
+        // over the static fallback that weFailed had already revealed
+        // underneath. That fallback exists to avoid exactly this.
+        onWeFailedChanged: if (bgRoot.weFailed) bgRoot.settleWeTransition()
+        // Same hole, different trigger: dropping weActive destroys the surface
+        // the peel samples as its destination, so the frozen still is all that
+        // is left to paint.
+        onWeActiveChanged: if (!bgRoot.weActive) bgRoot.settleWeTransition()
+
         // Switch the WE surface to `path`, animating a shader transition from a
         // snapshot of the current frame into the newly-loaded surface. Called on
         // weProjectPath changes instead of a reactive binding so the old frame can
@@ -353,11 +379,7 @@ Variants {
             running: bgRoot.weTransitioning
             onTriggered: {
                 console.warn("[Background] Wallpaper Engine transition did not finish; settling")
-                weTransitionAnim.stop()
-                weTransitionDelay.stop()
-                bgRoot.weTransitionProgress = 1.0
-                bgRoot.weTransitioning = false
-                weOldStill.source = ""
+                bgRoot.settleWeTransition()
             }
         }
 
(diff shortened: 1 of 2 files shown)


XephyLon:
## What changed

**`6ebb8c9` — #92**

The issue proposes invalidating the stash on `loadWeWallpaper`'s early returns. I removed the stash instead, and that is the one decision here worth pushing back on if you disagree.

```qml
onSuppressContentsChanged: {
    if (!bgRoot.suppressContents)
        bgRoot.loadWeWallpaper(bgRoot.weProjectPath);
}
```

Reasoning: invalidation has to cover every route that can supersede a stash, and the issue already documents two (the `weLoadedProject` early return, and `weActive` dropping so `onLoaded` re-syncs while the stash is untouched). `weProjectPath` is the request itself, so it cannot be stale by construction, and `loadWeWallpaper` already no-ops when the path matches what is loaded — the unconditional call is free. It also disposes of the `""` overload you flagged under "Also worth folding in", which invalidation alone would leave in place.

The cost is that `wePendingProject` is gone, so if you were planning to hang anything else on it, this is the wrong shape. Say so and I will redo it as the two-arm clear instead.

**`7e4aafe` — #94**

Both remedies you suggested, not one:

```qml
visible: bgRoot.weTransitioning && !bgRoot.weFailed && weLoader.item !== null
```

plus `onWeFailedChanged` / `onWeActiveChanged` calling a shared `settleWeTransition()`. The guard makes the peel's lifetime a property of the shader rather than of getting handler ordering right; the settle is what stops it waiting out a watchdog whose whole budget exists to cover a slow *healthy* load. The watchdog's own recovery was already correct — it just became the third caller of the same function so the three cannot drift.

## How I verified it

**Every pin was proven by mutation** — written, run against the broken code, seen to fail, reverted. Five mutations, all confirmed:

| mutation | pin that caught it |
|---|---|
| stash restored verbatim | `test_the_project_replayed_on_un_suppress_is_the_live_request` — `` replays `pending` rather than the live weProjectPath `` |
| deferral removed from `loadWeWallpaper` | `test_a_switch_made_while_suppressed_is_applied_on_the_way_back` |
| un-suppress replay removed | same, `un-suppressing must call loadWeWallpaper` |
| `visible: bgRoot.weTransitioning` restored | `test_the_transition_overlay_yields_to_the_static_fallback` — `'weFailed' not found in 'bgRoot.weTransitioning'` |
| both settle handlers deleted | `test_a_transition_that_can_never_finish_settles_without_the_watchdog`, both subtests |

Two existing pins had to change, and I want to be explicit that I did not loosen them:

- `test_a_switch_requested_while_suppressed_is_deferred` named `wePendingProject`, which no longer exists. Rewritten to pin the behaviour — declined while suppressed, applied on the way back — and it still fails on both halves being broken (rows 2 and 3 above). Its own docstring already said "Declaring `wePendingProject` is not the contract", so this is the pin it wanted to be.
- `test_transition_cannot_hang_forever` asserted `weTransitioning = false` inside the watchdog's `onTriggered`, which now delegates. It follows a call one level deep via a new `_with_called_bodies` helper, so gutting `settleWeTransition` still fails it.

I added `_QmlSource.function_body()` for that — QML function bodies are `js` blocks unreachable through `elements()`/`members()`.

**Suites:** `260 passed, 0 failed` throughout (both commits add Python pins, not QML tests; `test_background_fullscreen_suppression.py` went 19 → 22 tests). Green before each commit.

- **No live verification, and this path is unexercised anyway.** No Hyprland, no Wayland display, no `qs` in this environment. More to the point, the whole #94 failure path is unreachable on the installed pin: `7e58913` has no `failed` property, so `weFailed` is permanently `false` there. Nobody has produced an unrenderable wallpaper to watch this with. The renderer-side ordering (`failed` reported without `rendered` ever going true) is read off v0.2.0's `wallpaperenginesurface.cpp`, exactly as your issue says — I did not re-derive it and I did not observe it. **The ~9s figure remains reasoned, not measured.**
- **Did not touch `weTransitionWatchdog`'s interval.** With the two states settling immediately it now only covers genuinely slow loads, which is what its comment says it is for.
- **Did not make `settleWeTransition` stop the watchdog explicitly** — its `running: bgRoot.weTransitioning` binding already does, and adding a redundant `.stop()` would suggest the binding is not trusted.
- **Did not add a log line for either settle path.** The watchdog warns because reaching it means something went wrong unobserved; a failed project already surfaces through `weFailed` and the visible fallback, so a second warning would be noise. Reconsider if you would rather see it in `log.log`.
- **Did not touch the static-image transition path**, `lockPeel`, or `PluginWidget.weSurfaceItem` — the issue traced all of those as already falling back correctly, and I did not re-audit them.
- **`superpowers` was not available** in this session and there is no marketplace to install it from in this container. CONTRIBUTING.md makes it a hard requirement, so I am flagging it rather than proceeding silently.

---

Issue #97, opened 2026-08-03
Background suppression tests: the load-bearing PanelWindow-visible pin matches nothing and asserts nothing


XephyLon:
## What breaks

`tests/test_background_fullscreen_suppression.py` (added in 4e46d78dd) exists to stop one specific regression from coming back: binding the background `PanelWindow`'s own `visible` to a fullscreen condition, which under `WlrLayershell` destroys the layer surface rather than hiding it, and produced a 30 Hz full-screen strobe. The file's own docstring calls that "the pin that matters most".

That pin does not currently assert anything, and would not catch the most likely form of the regression.

### The load-bearing pin is vacuous today

`dots/.config/quickshell/imi/tests/test_background_fullscreen_suppression.py:34-48`

```python
window_level = re.findall(r"^        visible:.*$", self.background, re.M)
for line in window_level:
    self.assertNotRegex(line, r"[Ff]ullscreen", ...)
```

`Background.qml` has **zero** lines matching `^        visible:` — verified by running that exact regex against the current file. The list is empty, the loop body never executes, and the test passes without evaluating a single assertion. It is a no-op that reports green.

### It would also miss the natural way to reintroduce the bug

Even once a `visible:` does appear at that indent, the assertion only rejects the literal substring `fullscreen`/`Fullscreen` on the same physical line. Given how the file is written now, the obvious way someone re-breaks this is:

```qml
visible: !bgRoot.suppressContents      // passes: no "fullscreen" substring
```

`suppressContents` is the property the whole suppression mechanism is named after, and it is the exact expression already used on the content `Item` at `Background.qml:503` — so moving it up one level is a one-line edit that reads as a simplification and sails through. A multi-line binding, or any reindent of the file, also defeats it.

### Three more assertions check that identifiers exist, not that behaviour holds

- `:96-101` `test_a_switch_requested_while_suppressed_is_deferred` asserts only `assertIn("wePendingProject", self.background)`. Passes if the property is declared and never read. (It is in fact half-broken right now — see the companion issue on stale replay — and this test is green.)
- `:103-105` `test_transition_cannot_hang_forever` asserts only `assertIn("weTransitionWatchdog", ...)`. Passes if the `Timer` exists with `running: false` and no `onTriggered`.
[...]

PR #109, opened 2026-08-04
fix(sidebarRight): hide the empty-notification shape when the list is cramped (#87)

### LINKED ISSUE #87: Right sidebar: no notification icon not in container
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://ii.clsty.link/en/ii-qs/04troubleshooting/) and [Usage](https://ii.clsty.link/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://ii.clsty.link/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

...

### Step 3. Describe the issue

The big icon is the most obvious issue, but the other small containers also reach outside the big notifications container on the top.

Image

### Reminder

- [x] I agree that it's usually impossible for others to help me without my logs.

XephyLon:
## Describe your changes

Implements your call on #87 — *"I'll just hide the notification icon when covered by too many things."*

`PagePlaceholder` centres its column with `anchors.centerIn` on a page that does not clip, so a page shorter than the column draws it **outside** the container rather than cropping it. The right sidebar's notification list shares its column with `BottomWidgetGroup`'s fixed 352px, so on a short screen it is squeezed until the 56px shape reaches above the rounded container's top edge.

The shape is the tallest part of a placeholder and the part carrying the least information, so it is what gives way — `"Nothing"` on its own still reads as an empty state. **Opt-in** via `dropIconWhenCramped`, set only by the notification list; the other five placeholders are byte-for-byte unaffected.

The fit rule lives in `modules/common/functions/placeholderFit.js` (same shape as `screenSelection.js`) because the widget cannot be instantiated by the QML suite — `StyledText` assigns `font.variableAxes`, which is Qt 6.7, and the suite runs against the distribution Qt 6.4. `tst_placeholder_fit.qml` drives it directly; three pins in `test_shared_widget_contracts.py` keep that from going vacuous.

Ready. One thing I deliberately left alone is the reporter's *second* complaint — details in the review comment below.

Closes #87

### DIFF
--- a/dots/.config/quickshell/imi/modules/common/functions/placeholderFit.js
+++ b/dots/.config/quickshell/imi/modules/common/functions/placeholderFit.js
@@ -1,0 +1,31 @@
+.pragma library
+
+// A PagePlaceholder centres its column in whatever space the page gives it,
+// and a page shorter than that column does not clip it - the column simply
+// draws outside the container it belongs to. The right sidebar's notification
+// list is the case that forced this (#87): it shares the sidebar column with a
+// fixed-height bottom widget group, so on a short screen it is squeezed until
+// the placeholder's shape sits above the list's own top edge, outside the
+// rounded rectangle that is supposed to hold it.
+//
+// The shape is both the tallest part of a placeholder and the part carrying
+// the least information, so it is what gives way first - the text alone still
+// says what the empty state is.
+
+// Whether the shape still fits above the labels within `availableHeight`.
+//
+// `iconHeight` is the shape's own implicit height and `textHeights` are the
+// visible labels' - never the column's own implicit height, which excludes an
+// invisible child. Deciding from that would drop the shape, free its height,
+// find there is room again, put it back, and oscillate forever.
+//
+// A page with no height yet has not been laid out; it is not "too short". A
+// placeholder built before its first layout pass would otherwise start without
+// its shape and pop it in, which reads as a glitch rather than as a fit.
+function iconFits(availableHeight, iconHeight, textHeights, spacing) {
+    if (!(availableHeight > 0))
+        return true;
+    const text = (textHeights || []).reduce(
+        (total, height) => total + height + spacing, 0);
+    return availableHeight >= iconHeight + text;
+}
--- a/dots/.config/quickshell/imi/modules/common/widgets/PagePlaceholder.qml
+++ b/dots/.config/quickshell/imi/modules/common/widgets/PagePlaceholder.qml
@@ -2,6 +2,7 @@ import QtQuick
 import QtQuick.Layouts
 import qs.modules.common
 import qs.modules.common.widgets
+import "../functions/placeholderFit.js" as PlaceholderFit
 
 Item {
     id: root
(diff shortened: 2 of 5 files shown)


XephyLon:
## What changed

`PagePlaceholder` gains an opt-in `dropIconWhenCramped`; `NotificationList` is the only thing that sets it. Three decisions in here are less obvious than the diff makes them look:

**The measurement never reads the column's `implicitHeight`.** A layout's implicit height excludes an invisible child, so deciding from it would drop the shape, free its height, find there is room again, put it back — a relayout oscillation, which per AGENT.md means a pegged core rather than a wrong pixel. It sums `shapeWidget.implicitHeight` and the visible labels' own implicit heights instead, none of which move when the shape is hidden.

**The column takes an explicit width while measuring.** Same failure by the other axis: a description wraps to the column's width, so leaving that width to the children would let dropping the shape narrow the column, reflow the description, change the height being measured, and feed back into the decision. Only applies when opted in, so nothing else re-wraps.

**The placeholder is given the list's area, not the whole column.** It was filling the entire `NotificationList` item, status row included. Measuring those ~40px as free would keep a shape that overlaps the buttons. This does shift the empty state up by about half the status row's height — a deliberate change, and the one visual difference here that is not purely "shape present or absent".

Why opt-in rather than automatic for all six placeholders: the other five sit on pages with room to spare where the shape *is* the empty state, and I cannot look at any of them from here to confirm a threshold behaves. Say the word and I will flip the default.

## How I verified it

The widget itself cannot be instantiated by the QML suite. I confirmed that rather than assuming it — `StyledText` assigns `font.variableAxes`, and against this Qt:

```
FAIL!  : tst_probe::compile() Cannot assign to non-existent property "variableAxes"
```

It is Qt 6.7+, and both this container and CI's `apt` Qt are 6.4.2. Mocking around it does not work either: `font` is special-cased by the QML engine, so a custom grouped `font` property on a stub is shadowed by the built-in value type. So the decision moved into `placeholderFit.js` and is exercised for real, in the same shape as `screenSelection.js` / `monitorDetection.js`.

**Six mutations, all confirmed failing before landing:**

| # | mutation | caught by |
|---|---|---|
| 1 | `>` instead of `>=` | `test_theBoundaryIsExactFit` (+2 others) — `an exact fit still fits` |
| 2 | per-label spacing dropped from the sum | `test_theBoundaryIsExactFit` — `one pixel short does not` |
| 3 | not-yet-laid-out guard removed | `test_anUnlaidOutPageIsNotTreatedAsTooShort` |
| 4 | shape no longer gated on `iconShown` | `test_the_placeholder_shape_is_gated_on_the_tested_fit_rule` |
| 5 | measure `column.implicitHeight` instead | `test_the_placeholder_measures_the_children_not_the_column` |
| 6 | `dropIconWhenCramped` / list area removed | `test_the_notification_placeholder_opts_in_and_gets_the_lists_area` |

Mutations 1–3 are the arithmetic; 4–6 are the wiring, and exist because a unit test on a library function goes silently vacuous the moment the widget stops calling it — the failure mode #97 was about.

**Suites:** `260 passed, 0 failed` → `268 / 0` (8 new QML cases; `test_shared_widget_contracts.py` went 1 → 4). Green before the commit.

- **I could not see the screenshots.** Both attachments 403 through this session's proxy (`sessions are bound to their configured repositories`). Everything here is derived from the code — `CenterWidgetGroup` is a `Rectangle` with no `clip`, `PagePlaceholder` centres on a non-clipping page, `BottomWidgetGroup.expandedHeight` is a fixed 352 that the list pays for. That is a complete account of *"the big icon … reaches outside the big notifications container on the top"*, but **I never saw the symptom**, and I have not seen the fix either.
- **Did not address "the other small containers also reach outside".** You called that cramming rather than a bug and scoped the fix to the icon, so I did not go looking for it. If it is still visible after this, it is a separate change — most likely `clip: true` on `CenterWidgetGroup`, which I left alone deliberately: clipping hides overflow rather than resolving it, and it would silently crop anything else that legitimately extends past that rectangle.
- **Did not give `CenterWidgetGroup` a `Layout.minimumHeight`.** That is the structural fix — the list can currently be squeezed toward zero — but it takes height from the widget group or the media player, which is a layout-priority decision that is yours, not a bugfix.
- **Did not touch `BottomWidgetGroup.expandedHeight`.** AGENT.md is explicit that the constant is load-bearing and that content-sizing it quietly costs the notification list ~36px.
- **No live verification.** No Hyprland, no Wayland display, no `qs` here, so neither the log loop nor `DesignSystemCompile` ran, and `PagePlaceholder` is one of the widgets the QML suite provably cannot compile — this is exactly the class AGENT.md warns stays green while failing to build. **Worth opening the right sidebar with no notifications before merge**, on a short screen and a tall one.
- **`superpowers` was not available** in this session and there is no marketplace to install it from in this container. Flagging per CONTRIBUTING rather than proceeding silently.

---

PR #110, opened 2026-08-04
Reboot after arming BootNext, and disarm it if the reboot fails

### LINKED ISSUE #104: "Reboot Into" arms BootNext and then never reboots, leaving the machine silently set to boot another OS
## What breaks

Picking an entry under **Reboot Into** in the session screen sets the firmware's `BootNext` and then does not reboot. The user sees a polkit password prompt, authenticates, and nothing else happens — so it reads as "the button does nothing".

It is worse than doing nothing: `BootNext` is a one-shot firmware flag that persists until consumed, so the machine is left silently configured to boot a **different OS** on its next restart, whenever that happens.

## Evidence from a live machine

After one failed attempt at "Reboot Into → Windows Boot Manager", with no reboot having occurred:

```
$ efibootmgr
BootNext: 0000        <- set by the attempt, still armed
BootCurrent: 0001     <- still running arch
BootOrder: 0001,0000,0003,0004,0002,0005,0006,0007
Boot0000* Windows Boot Manager
Boot0001* arch
```

So the privileged half **succeeded**: `pkexec efibootmgr -n 0000` ran, authenticated and wrote the flag. Only the reboot is missing.

Every binary in the chain is present: `efibootmgr`, `pkexec`, `reboot`, `loginctl`, `systemctl` all resolve.

## The chain

`modules/imi/sessionScreen/SessionScreen.qml:310` → `EfiBoot.rebootInto(num, label)`
→ `services/EfiBoot.qml:60-64` sets `pendingNum` and starts `setNextProc`
→ `services/EfiBoot.qml:80` `command: ["pkexec", "efibootmgr", "-n", root.pendingNum]`
→ on `exitCode === 0`, `services/EfiBoot.qml:83` calls `Session.reboot()`
→ `modules/common/functions/Session.qml:45-48`:

```qml
function reboot() {
    closeAllWindows();
    Quickshell.execDetached(["bash", "-c", `reboot || loginctl reboot`]);
}
```

## Leading suspicion, not yet proven

`closeAllWindows()` runs **before** the detached reboot and tears down the shell's own windows — the session screen among them. If that teardown takes the process (or the detached child) with it, `execDetached` never lands, which matches the symptom exactly.

Worth checking as part of the same investigation:

- whether plain `reboot` succeeds for an unprivileged user in this session (logind/polkit `org.freedesktop.login1.reboot`), since the `|| loginctl reboot` fallback only helps if the first form fails *fast* rather than hanging;
- whether the ordinary **Reboot** button in the same session screen works — it calls the same `Session.reboot()`. If it also fails, this is not EFI-specific at all and the title is too narrow.

## Fix direction

Reboot first, or at minimum do not tear down the UI until the reboot call has been issued and seen to succeed. And if the reboot cannot be issued, **unset `BootNext`** (`efibootmgr -N`) rather than leaving the machine armed for an OS switch the user did not get — a failure here should not persist past the failed action.

## Severity

**High.** Silent, user-visible, and it leaves persistent state the user has no reason to suspect: the next reboot — days later, for unrelated reasons — boots the wrong OS.

XephyLon:
## Describe your changes

Closes #104.

The cause is an **unresolved identifier, not a race.** `EfiBoot.qml` called `Session.reboot()` while importing only `qs.modules.common` and `qs`. `Session` is declared in `qs.modules.common.functions`, QML imports are not transitive, so the bareword never resolved — the call threw `ReferenceError` inside `onExited` and the handler stopped there. `pkexec` had already succeeded, so `BootNext` stayed armed, and no notification fired because the existing `else if` only reports a *non-zero* exit while the exit code was `0`.

The issue's leading suspicion — `closeAllWindows()` tearing down the shell before `execDetached` lands — **does not hold**; evidence in the comment below.

Four commits:

1. **`fix(efiboot)`** — the reboot becomes a `Process` with an observable exit code, deliberately not `Session.reboot()`. Once `BootNext` is armed, "the reboot silently did not happen" is exactly the state that must not go unnoticed, and `execDetached` reports nothing back, so there is no way to know a rollback is needed. On a non-zero exit the `BootNext` is cleared with `pkexec efibootmgr -N`; if that also fails, the notification names the entry that will boot and the exact command to fix it.
2. **`fix(session)`** — separate defect found on the way. `reboot()` / `poweroff()` / `rebootToFirmware()` killed every one of the user's windows *before* issuing the transition, so a transition that failed left an empty desktop, still logged in, with nothing explaining why. Order reversed.
3. **`test(lint)`** — the import lint that should have caught this knew only about `Appearance` and walked only `modules/`. It missed `EfiBoot.qml` on both axes. Now table-driven and scans `services/`.
4. **`test(efiboot)`** — extends `test_efiboot_contract.py` (7 new cases, 14 total).

### Verification

Root cause confirmed under `qmltestrunner`, not by reading. Every new assertion proven by mutation — ten breakages, each seen to fail, each reverted. Full suite **260 passed, 0 failed** before each commit.

Ready to review. Not merged, nothing tagged, `WE_REF`/`SDDM_REF` untouched.

**One thing needs you on real hardware:** I could not test that the reboot now actually happens — that requires an actual reboot of your live desktop. Everything here is static analysis plus runtime QML checks. Details of exactly what remains unconfirmed are in the comment below.

### DIFF
--- a/dots/.config/quickshell/imi/modules/common/functions/Session.qml
+++ b/dots/.config/quickshell/imi/modules/common/functions/Session.qml
@@ -37,18 +37,29 @@ Singleton {
         Quickshell.execDetached(["bash", "-c", `systemctl hibernate || loginctl hibernate`]);
     }
 
+    // Order is deliberate: issue the transition FIRST, close windows after. It
+    // used to be the other way round, which meant a transition that did not
+    // happen - a polkit denial, an inhibitor, or a session logind does not
+    // consider "active" and therefore wants admin auth for - had already killed
+    // every one of the user's applications by the time it failed. The user was
+    // left on an empty desktop, still logged in, with nothing on screen
+    // explaining why.
+    //
+    // Reversing it costs nothing. systemd's shutdown sequence SIGTERMs every
+    // process anyway, so applications still get their chance to save;
+    // closeAllWindows() only ever brought that forward by a fraction of a second.
     function poweroff() {
-        closeAllWindows();
         Quickshell.execDetached(["bash", "-c", `systemctl poweroff || loginctl poweroff`]);
+        closeAllWindows();
     }
 
     function reboot() {
-        closeAllWindows();
         Quickshell.execDetached(["bash", "-c", `reboot || loginctl reboot`]);
+        closeAllWindows();
     }
 
     function rebootToFirmware() {
-        closeAllWindows();
         Quickshell.execDetached(["bash", "-c", `systemctl reboot --firmware-setup || loginctl reboot --firmware-setup`]);
+        closeAllWindows();
     }
 }
--- a/dots/.config/quickshell/imi/services/EfiBoot.qml
+++ b/dots/.config/quickshell/imi/services/EfiBoot.qml
@@ -75,17 +75,63 @@ Singleton {
         }
     }
 
+    function notify(title, body) {
+        Quickshell.execDetached(["notify-send", title, body, "-a", "Session"])
+    }
+
     Process {
         id: setNextProc
         command: ["pkexec", "efibootmgr", "-n", root.pendingNum]
         onExited: (exitCode, exitStatus) => {
             if (exitCode === 0) {
-                Session.reboot()
+                // BootNext is now armed in firmware. From here every exit path
+                // must either reboot or clear it again - see rebootProc.
+                rebootProc.running = true
             } else if (exitCode !== 126 && exitCode !== 127) {
                 // 126/127 = polkit dismissal/authorization failure: user's call,
                 // stay quiet. Anything else is a real failure worth surfacing.
-                Quickshell.execDetached(["notify-send", "Reboot cancelled",
-                    `efibootmgr failed to set BootNext (${exitCode})`, "-a", "Session"])
+                root.notify("Reboot cancelled",
+                    `efibootmgr failed to set BootNext (${exitCode})`)
+            }
+        }
+    }
+
+    // Deliberately a Process rather than Session.reboot(). Session.reboot() is
+    // execDetached, which reports nothing back - and once BootNext is armed,
+    // "the reboot silently did not happen" is precisely the state that must not
+    // go unnoticed, because it leaves the machine set to boot another OS at some
+    // unrelated future restart. An exit code is the whole point.
+    //
+    // Nothing tears down the UI first. The session screen has already hidden
+    // itself before rebootInto() is called, and a reboot that fails must leave
+    // the user's session exactly as it was.
+    Process {
+        id: rebootProc
+        command: ["bash", "-c", "reboot || loginctl reboot"]
+        onExited: (exitCode, exitStatus) => {
+            if (exitCode === 0) return // system is going down; nothing left to do
+            // The reboot definitively failed, so the armed BootNext is now a
+            // trap: the next restart, for whatever unrelated reason days later,
+            // would boot the other OS. Clear it. This needs privilege again and
+            // therefore a second polkit prompt - unavoidable, but it only
+            // happens on a path where something has already gone wrong.
+            disarmProc.running = true
+        }
+    }
+
+    Process {
+        id: disarmProc
+        command: ["pkexec", "efibootmgr", "-N"]
+        onExited: (exitCode, exitStatus) => {
+            if (exitCode === 0) {
+                root.notify("Reboot failed",
+                    `Could not reboot into ${root.pendingLabel}. BootNext was cleared, so this machine still boots as usual.`)
+            } else {
+                // Worst case: armed, could not reboot, could not disarm. The
+                // user has to know the exact command, because nothing on screen
+                // will hint that the next restart boots a different OS.
+                root.notify("Reboot failed - BootNext still set",
+                    `Could not reboot into ${root.pendingLabel}, and clearing BootNext failed (${exitCode}). The next restart will boot ${root.pendingLabel}. Run: sudo efibootmgr -N`)
             }
         }
     }
--- a/dots/.config/quickshell/imi/tests/lint_qml_imports.sh
+++ b/dots/.config/quickshell/imi/tests/lint_qml_imports.sh
@@ -1,44 +1,79 @@
 #!/usr/bin/env bash
 #
-# Regression guard: a .qml that uses the `Appearance` singleton as a bareword
-# (Appearance.colors, Appearance.spacing, ...) MUST import qs.modules.common,
-# where the singleton is declared. The import is NOT transitive through
-# qs.modules.common.widgets. Omitting it throws "ReferenceError: Appearance is
-# not defined" on every binding evaluation; when the missing token feeds a
-# positioner's spacing/margin, the resulting undefined -> NaN geometry thrashes
-# relayout and pegs a core at 100% CPU (see AGENT.md at the repo root). The appearance-token
-# migration introduced exactly this in several files, so this check keeps it
-# from coming back.
+# Regression guard: a .qml that uses a singleton as a bareword MUST import the
+# module that singleton is declared in. QML imports are NOT transitive -
+# `import qs.modules.common` does not bring in `qs.modules.common.functions` -
+# and a missing one is not a load-time error. The bareword simply stays
+# unresolved, and every use throws "ReferenceError: <X> is not defined" when it
+# is evaluated. Inside a signal handler that means the handler silently does
+# nothing at all.
 #
-# Exits non-zero and lists offenders if any file references bareword
-# `Appearance.` without an (unaliased) `import qs.modules.common`.
+# Two real bugs of exactly this shape:
+#
+#   Appearance - the appearance-token migration dropped `import qs.modules.common`
+#                from several files. The undefined token feeds a positioner's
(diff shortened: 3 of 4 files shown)


XephyLon:
## What actually caused it

An unresolved identifier. Not a teardown race.

`EfiBoot.qml` imported `QtQuick`, `Quickshell`, `Quickshell.Io`, `qs.modules.common`, `qs` — and called `Session.reboot()`. `Session` is declared in `qs.modules.common.functions`:

```
$ cat tests/imports/qs/modules/common/functions/qmldir
module qs.modules.common.functions
...
singleton Session 1.0 Session.qml
```

`qs.modules.common`'s own qmldir declares `Appearance`, `Config`, `Directories`, `Icons`, `Images`, `Persistent` — **not** `Session`. QML imports are not transitive, so the bareword never resolved and `Session.reboot()` threw `ReferenceError` inside `onExited`.

Everything about the symptom follows from that:

- `pkexec efibootmgr -n 0000` had **already** succeeded, so `BootNext` stayed armed — matching the `efibootmgr` output in the issue exactly.
- No notification, because the existing `else if (exitCode !== 126 && exitCode !== 127)` only fires on a **non-zero** exit and the exit code was **0**. The success path was the silent one.
- Every other file calling `Session.*` — `SessionScreen.qml`, `LockScreen.qml`, `LockSurface.qml`, `NotificationItem.qml`, `NotificationPopupItem.qml` — imports `qs.modules.common.functions`. `EfiBoot.qml` was the only one that did not.

## How I established it without rebooting

**Runtime, under `qmltestrunner`, against the repo's own mock import tree** — a throwaway `TestCase` with EfiBoot.qml's exact import set:

```
EfiBoot import set    -> typeof Session = undefined
FAIL!  : 'Session unresolved under EfiBoot.qml's imports' returned FALSE
```

and the control, with the one missing import added:

```
with .functions added -> typeof Session = object
PASS   : test_b_with_functions_import()      # Session.reboot is a function
```

That is the whole bug, demonstrated in both directions, with nothing rebooted.

**The stated suspicion is disproven.** `closeAllWindows()` maps over `HyprlandData.windowList`, which is filled from `hyprctl clients -j` — xdg toplevels. The session screen is a `PanelWindow` with `WlrLayershell.namespace: "quickshell:session"`, and layer surfaces come from a *separate* `hyprctl layers -j` call (`HyprlandData.qml:169`) that never feeds `windowList`. So `closeAllWindows()` cannot kill the shell or the session screen. `SessionScreen.qml:309` also already calls `sessionRoot.hide()` **before** `EfiBoot.rebootInto(...)`, so the UI is gone before `pkexec` even starts. There was no race to lose.

**Can an unprivileged user reboot?** Yes, in an active session. From systemd's own shipped policy file (package data, identical on Arch):

```
org.freedesktop.login1.reboot
   allow_any        = auth_admin_keep
   allow_inactive   = auth_admin_keep
   allow_active     = yes
```

So `reboot` succeeds without authentication for an active local session, returns fast, and `|| loginctl reboot` is not normally reached. It also means the reboot *would* have worked had it been called — further isolating the fault to the unresolved identifier. Caveat: I read the policy defaults, and I could not check your machine for local `.rules` overrides (this container has no running logind).

## Does the plain Reboot button share the bug?

**No — but it shares the shape, and I fixed that too.**

It does not hit the `ReferenceError`: `SessionScreen.qml` imports `qs.modules.common.functions`, so `Session.reboot()` resolves there. The issue title is not too narrow on that count.

But `Session.reboot()`, `poweroff()` and `rebootToFirmware()` all called `closeAllWindows()` **first** and issued the transition second. If the transition then failed — polkit denial, a shutdown inhibitor, or a session logind does not consider "active" and so wants admin auth for — every one of the user's applications had already been SIGTERMed. Empty desktop, still logged in, nothing on screen explaining it. Same family as #104: a failed action leaving destructive state behind.

Order reversed. Nothing is lost: systemd's shutdown SIGTERMs everything anyway, so apps still get to save; `closeAllWindows()` only ever brought that forward by a fraction of a second. On the failure path it is now the difference between an empty desktop and an untouched one.

`logout()` is deliberately **not** reordered — `pkill -i Hyprland` has no polkit-gated failure path that could strand the user. That is pinned by a test so a later consistency pass does not change it without the argument.

## What the fix does when the reboot cannot be issued

`pkexec efibootmgr -N`, clearing the armed `BootNext`, then a notification.

The reboot is now a `Process` rather than `Session.reboot()`, and that is the load-bearing part. `execDetached` returns nothing — no exit code, so "the reboot never happened" is undetectable and rollback is impossible. Owning it as a `Process` is what makes the rollback expressible at all.

Three outcomes:

| outcome | behaviour |
|---|---|
| reboot exits 0 | system is going down; nothing further, and **no** disarm — that would clear the `BootNext` the user just asked for, on the way out |
| reboot exits non-zero | `pkexec efibootmgr -N`, then *"Could not reboot into X. BootNext was cleared, so this machine still boots as usual."* |
| disarm also fails | *"...The next restart will boot X. Run: `sudo efibootmgr -N`"* — names the entry and the exact command, because nothing else on screen would ever hint at it |

The disarm is a second polkit prompt. That is unavoidable — clearing `BootNext` is privileged and the first authorization is long since consumed — but it only happens on a path where something has already gone wrong, and its own failure is surfaced rather than swallowed. Polkit dismissal (126/127) on the initial arm stays silent, as before.

## Systemic guard

`lint_qml_imports.sh` exists for exactly this bug class and did not catch it — it knew only about `Appearance` and walked only `modules/`, while `EfiBoot.qml` uses `Session` and lives in `services/`. It is now table-driven (`singleton:module`) and scans `services/` as well.

Verified against the real defect, not a synthetic one: with `services/EfiBoot.qml` reverted to its pre-fix content the lint exits 1 and names the file; restore the fix and it passes.

It also now strips comments before matching. The fix carries a comment explaining why it does *not* call `Session.reboot()`, and a lint that reports a comment as a violation is one that gets switched off the first time it fires.

## Mutations

Every new assertion was proven by breaking the code and watching it fail. Control: clean tree passes.

| # | mutation | killed by |
|---|---|---|
| 1 | restore `Session.reboot()` — **the original bug verbatim** | `reboot_is_observable_not_fire_and_forget` |
| 2 | rename away the reboot `Process` | `reboot_is_observable_not_fire_and_forget` |
| 3 | never disarm on failed reboot | `failed_reboot_clears_bootnext` |
| 4 | disarm unconditionally (would kill a *good* reboot) | `failed_reboot_clears_bootnext` |
| 5 | drop the `pkexec efibootmgr -N` command | `failed_reboot_clears_bootnext` |
| 6 | disarm-failure message omits the command | `disarm_failure_names_the_manual_command` |
| 7 | EfiBoot's reboot command drifts from `Session`'s | `reboot_command_matches_the_plain_reboot_button` |
| 8 | `Session.reboot()` closes windows first again | `transition_is_issued_before_windows_are_closed` (`function='reboot'`) |
| 9 | lint drops its `Session` entry | `import_lint_still_covers_session_and_services` |
| 10 | lint stops scanning `services/` | `import_lint_still_covers_session_and_services` |

**All 10 fail; tree verified clean afterwards.** The harness aborts a mutation whose `sed` did not apply, so a no-op edit cannot be misread as a pass — that guard fired on mutation 8's first attempt and I re-ran it as a real line swap rather than recording a false result.

Worth noting what mutation 1 means: the old `test_bootnext_via_pkexec_then_reboot` asserted the **string** `"Session.reboot()"` was present. That assertion passed for the entire life of this bug — the text was there, the identifier was unresolved, nothing rebooted. It is now replaced by one that fails.

## Full suite

`cd dots/.config/quickshell/imi/tests && bash run_tests.sh`, before each of the four commits:

```
QML import lint passed: all singleton users import the declaring module
Running EFI boot contract tests...  Ran 14 tests  OK
Totals: 260 passed, 0 failed, 0 skipped, 0 blacklisted
All tests passed successfully!    EXIT=0
```

## Still needs a human on real hardware

**Whether the machine now actually reboots.** This is the one claim I cannot make. Everything above is static analysis plus runtime QML resolution checks; no reboot was performed, because it is your live desktop and you are away. The evidence says the call now resolves and is issued, and that logind permits it for an active session — but "resolves and is issued" is not "the machine came back up in Windows". Please try it once.

Two narrower items:

- **Your polkit ruleset.** I read systemd's shipped defaults from package data. Local `/etc/polkit-1/rules.d/*.rules` could override `org.freedesktop.login1.reboot`; I had no running logind to query with `pkcheck`.
- **The rollback path has never executed.** Exercising it needs a reboot that genuinely fails — hard to induce deliberately without a shutdown inhibitor. The wiring is pinned by tests and the commands are two flags on a binary already proven to work on your machine (`efibootmgr -n` succeeded), but the notification text has not been seen on a real screen.

One residual limitation, recorded rather than papered over: if the reboot command **hangs** instead of returning non-zero — a polkit prompt in a non-active session, say — `onExited` never fires, so nothing disarms and the machine stays armed. I considered a timeout that disarms after N seconds and rejected it: `systemctl reboot` returns 0 as soon as the transition is queued, but a slow shutdown can take 90s, and a timeout firing during a genuine reboot would clear the `BootNext` on the way down and boot the wrong OS — turning a rare hang into a broken feature. Rolling back only on an explicit non-zero exit is the safe half of that trade.


---

PR #111, opened 2026-08-04
feat(weather): day-by-day forecast in the bar popup (#81)

### LINKED ISSUE #81: forecast for weather
### What would you like to be added?

I would like to see forecast added to the weather widgets.

### How will it help?

Right now, I can only see the current weather, which is already nice, but forecast would make the widgets even more useful since I'd be able to use them for planning when to go out etc.

### Extra info

There are many shells that have this feature, for example:
https://raw.githubusercontent.com/na-ive/nandoroid-docs/refs/heads/main/public/showcase2.png
https://danklinux.com/

XephyLon:
## Describe your changes

Adds the forecast requested in #81, as a row of day cards in the bar weather popup.

`Weather.forecast` is `[{ date, wCode, high, low }]` — a local `YYYY-MM-DD` and whole degrees in the configured unit system, identical for both providers.

- **wttr.in costs nothing.** `?format=j1` already returns three days in `weather[]`; the service was fetching that and reading only `weather[0]`.
- **OpenWeatherMap needs a second request** to `/data/2.5/forecast` (free tier, 5 days / 3 hours), grouped into days here. So the reporter's open-meteo suggestion isn't needed — both existing providers can do it.

Along the way it fixes a real bug the feature would otherwise have inherited: the popup resolved OWM condition ids through the WWO-keyed `weatherIconMap`, drawing a clear sky through every storm for anyone on the default provider.

Three commits: provider-aware icons → forecast in the service → forecast row in the popup.

Ready. Two things need your judgement — the OWM call-rate increase, and the desktop weather card being deliberately out of scope. Both in the review comment below.

Closes #81

### DIFF
--- a/dots/.config/quickshell/imi/modules/common/Icons.qml
+++ b/dots/.config/quickshell/imi/modules/common/Icons.qml
@@ -95,4 +95,36 @@ Singleton {
         }
         return isNight() ? "clear_night" : "clear_day";
     }
+
+    // OpenWeatherMap reports its own condition ids, not the WWO codes
+    // weatherIconMap is keyed on, so an OWM id looked up there matches nothing
+    // and falls through to "clear" for every condition there is. The groups
+    // below are OWM's documented ranges; 800 is the only exact id it defines,
+    // and 801-804 are increasing cloud cover.
+    function getOwmWeatherIcon(code, night): string {
+        const id = Number(code);
+        if (id === 800) return night ? "clear_night" : "clear_day";
+        if (id === 801 || id === 802) return night ? "partly_cloudy_night" : "partly_cloudy_day";
+        if (id > 802 && id < 900) return "cloud";
+        if (id >= 200 && id < 300) return "thunderstorm";
+        if (id >= 300 && id < 400) return "rainy";
+        if (id >= 500 && id < 600) return "rainy";
+        if (id >= 600 && id < 700) return "snowing";
+        if (id >= 700 && id < 800) return "foggy";
+        return night ? "clear_night" : "clear_day";
+    }
+
+    // Which of the two schemes a code is in depends on which provider fetched
+    // it, and nothing in the code itself says - the ranges overlap. Callers that
+    // know their provider should come through here rather than guess.
+    function getProviderWeatherIcon(provider, code, night): string {
+        if (provider === "owm")
+            return getOwmWeatherIcon(code, night);
+        const key = String(code);
+        if (weatherIconMap.hasOwnProperty(key)) {
+            const icons = weatherIconMap[key];
+            return night ? icons.night : icons.day;
+        }
+        return night ? "clear_night" : "clear_day";
+    }
 }
--- a/dots/.config/quickshell/imi/modules/common/functions/weatherForecast.js
+++ b/dots/.config/quickshell/imi/modules/common/functions/weatherForecast.js
@@ -1,0 +1,120 @@
+.pragma library
+
+// Neither weather provider describes a forecast as days, so turning one into a
+// row of day cards is arithmetic over a flat list rather than a field lookup.
+//
+// wttr.in's `?format=j1` response already carries three days in `weather[]`,
+// each with its own min/max and eight three-hourly entries - the forecast comes
+// free with the request the service already makes. OpenWeatherMap splits it:
+// /data/2.5/weather is current conditions only, and /data/2.5/forecast is a
+// flat list of three-hourly entries that has to be grouped into days here.
+//
+// Both shapes collapse to the same `{ date, wCode, high, low }`, with `date` as
+// a local "YYYY-MM-DD" so the caller can label it and temperatures already in
+// whichever unit system was asked for.
+
+// Four fits the popup's width at both providers' resolutions. wttr.in only ever
+// returns three days, so that is what it shows; OWM's five-day list is cut here.
+const MAX_DAYS = 4;
+
+// The hour a day's icon is taken from. Not the first entry of the day: a
+// forecast keyed off the 00:00 observation reads as clear-and-cold for a day
+// that is in fact raining from lunchtime on, which is the opposite of what the
+// row is for.
+const ICON_HOUR = 12;
+
+// Local calendar date as "YYYY-MM-DD".
+//
+// Deliberately not `toISOString().slice(0, 10)`, which is UTC: anywhere east or
+// west of it that flips "today" hours early or late, and the row would put
+// today's label on tomorrow's card for part of every day.
+function localIsoDate(date) {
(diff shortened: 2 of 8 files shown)


XephyLon:
## What changed

**`4df8054` — provider-aware icons.** Not scope creep: the popup resolved OpenWeatherMap condition ids through `weatherIconMap`, which is keyed on the WWO codes wttr.in speaks. An OWM id misses and falls through to the clear-sky glyph, so on the *default* provider the popup drew a clear sky through every storm. Shipping a forecast row with correct icons above a current-conditions icon that is always wrong wasn't an option.

Falling through is only the loud half. The schemes overlap numerically, so some OWM ids land on an unrelated WWO entry and come back confidently wrong — OWM 302 is heavy drizzle, WWO 302 is drawn as hail. No "is it showing clear?" check would catch that, so it is pinned by name. OWM 200 and WWO 200 both mean thunderstorm, which is luck, and that is pinned as luck.

**`c325725` — the forecast.** Two decisions that separate a useful row from a misleading one:

- **A day's icon comes from the entry nearest local midday**, not the first of the day. Keyed off the 00:00 observation, a day that rains from lunchtime on renders clear and cold.
- **A day's range is the extremes across all its entries.** Each OWM entry's `temp_min`/`temp_max` describes its own three-hour window; reading either off one entry reports that window's range as the day's.

**`501b5c0` — the row.** Hidden rather than blank when there is no forecast (OWM's is a separate request that can fail on its own). Cards ask for the day icon explicitly — after 20:00 every card would otherwise flip to its night variant, and Thursday's forecast is not a claim about tonight's sky.

## How I verified it

**15 mutations, every one confirmed failing before landing.** Six on the forecast arithmetic, two on the icon ranges, seven on the wiring.

One of them is the reason I want to flag this section specifically.

> **A mutation survived, and that is what the timezone check is for.**
>
> Replacing `localIsoDate` with `date.toISOString().slice(0, 10)` — the exact UTC bug the function is written to avoid — left `tst_weather_forecast.qml` **fully green**:
>
> ```
> === M4: localIsoDate via toISOString (UTC) ===
> Totals: 15 passed, 0 failed
> ```
>
> Because this container and CI both run on UTC, where local and UTC are the same string for every instant. The test was real, but its coverage of that function was ambient — it would have shipped a pin that could never fire, which is precisely the failure mode of the earlier patch that passed 13/13 while its bug sailed through.
>
> `test_weather_forecast_contract.py` now re-runs that file under `Pacific/Kiritimati` (UTC+14) and `Pacific/Niue` (UTC-11). With the mutation reapplied:
>
> ```
> FAIL: test_the_forecast_unit_test_passes_east_and_west_of_utc (timezone='Pacific/Kiritimati')
>   FAIL!  : test_localIsoDateIsLocalNotUtc() single-digit months and days are padded
>      Actual   (): 2026-01-08
>      Expected (): 2026-01-09
> FAIL: ... (timezone='Pacific/Niue')
> ```
>
> Fails in both directions now. It also refuses to skip if `qmltestrunner` is missing, rather than passing quietly.

The other fourteen, in brief:

| area | mutation → pin that caught it |
|---|---|
| forecast | day range read off one entry; icon from midnight; wttr `time` read as an hour not hhmm; null rounded to 0°; wttr ignoring `useUSCS` |
| icons | `getProviderWeatherIcon` not branching on provider (5 cases fail); cloud-cover boundary shifted |
| wiring | forecast request never issued; **URL spliced into `sh -c "curl …"`**; `units=` dropped; wttr forecast never parsed; row shown blank; popup reverted to `getWeatherIcon`; cards taking the night variant |

The injection one matters: `city` reaches these URLs from config and from shareable presets, and this file carries a comment recording that it was hardened against exactly that once already. The pin checks *every* curl command in the service, not just the new one.

**Suites:** `260 passed, 0 failed` → `270 / 0` after commit 1 → `285 / 0` after commits 2 and 3. Green before each commit, never below baseline.

- **The desktop weather plugin card is untouched**, and this is the gap between what #81 asked for and what this delivers — the reporter's showcase links are desktop widgets. `DesktopWeatherWidget` is a fixed 108px single-row card with three *width-only* size modes and a horizontal resize handle; a forecast needs a new taller mode, which means a layout design decision plus grid-lattice, size-mode and settings plumbing, and `test_expressive_design_system.py`'s expectations. That is a feature of its own, and not one I can design blind. Happy to do it as a follow-up if you say what the taller card should look like.
- **No live verification.** No Hyprland, no Wayland display, no `qs`, and no network for a real API response — every provider fixture here is hand-built from the documented schemas. The forecast row has never been rendered. `qmllint` shows no syntax errors and the JS is exercised for real, but per AGENT.md that is exactly the class of change that stays green while failing to compile. **Worth hovering the bar weather widget before merge**, on both providers.
- **The OWM call rate doubles** — one extra request per `fetchInterval`, ~288/day at the 10 minute default. Fine against the free tier's 1M/month, but it is shared with the built-in fallback key, so it is your call. If you would rather not, the alternative is fetching the forecast only while the popup is open.
- **No new config option**, per CONTRIBUTING. The row appears when there is a forecast and hides when there is not. Say if you want it behind a toggle.
- **Did not touch the existing `getWeatherIcon`.** Its clock-reading behaviour is what its other callers expect; the new entry point takes the night flag as an argument instead.
- **Did not unify the icon vocabularies.** The desktop card has its own OWM range table because it renders `CustomIcon`s from the google-weather asset set — a different vocabulary from the Material Symbol names this returns, so sharing them is a port, not a rename.
- **Did not cap wttr and OWM to the same day count.** wttr gives 3, OWM 4 after capping. Padding to a fixed count means a placeholder card for a day nobody has data on.
- **`superpowers` was not available** in this session and there is no marketplace to install it from in this container. Flagging per CONTRIBUTING rather than proceeding silently.

---

XephyLon:
Closing: the bar popup is not going to carry a forecast row — there is nowhere to put one without spoiling that bar's look.

The other two thirds of this PR are landed, unchanged, as #190:

- `fix(weather): resolve icons against the provider that reported the code` — a real bug: OWM condition ids resolved through the WWO-keyed map drew a clear sky through every storm on the default provider.
- `feat(weather): publish a day-by-day forecast from both providers` — `Weather.forecast`, including the local-midday icon choice, the across-all-entries day range, and the timezone pin that re-runs the unit tests east and west of UTC.

The forecast itself now has a home on the desktop instead: #189 puts it in a 3x2 weather card, verified against live OpenWeatherMap data. The weekday fix this PR needed (`shortDayName` returned "8/14/26" where it meant "Fri", because QtQml replaces `toLocaleDateString` with a two-argument overload) went with it.

Only the popup row dies here. Nothing else in this work is lost.

Issue #113, opened 2026-08-04
Greeter background is capped at the Workshop preview's resolution (1920x1080 upscaled to 5120x1440)


XephyLon:
v0.14.9 pointed the greeter and the preset cards at `activePreview` instead of the dead `activeStill` field (#103). That was the right fix — `activePreview` is written on every wallpaper switch and always names the active project — but it capped the greeter background at whatever resolution the Steam Workshop preview happens to be.

The changelog for 0.14.9 says "Restoring proper still generation is tracked in #103". #103 is closed, so it isn't tracked anywhere. This is that tracker.

## Measured

On a 5120x1440 display, with a Wallpaper Engine **scene** project active:

| | source | resolution |
|---|---|---|
| desktop | live WE render | 5120x1440, native |
| greeter | `preview.gif` -> `Backgrounds/background.png` | **1920x1080**, upscaled 2.67x |

The theme itself is not the constraint — it ships a 7680x2160 `placeholder.png`. The constraint is that the only still we can hand it is a workshop preview.

The wider the display, the worse this gets. It is clearly visible at 5120x1440 and would be mild at 1920x1080.

## Why there is no better source today

The selector-only refactor removed the external renderer that used to produce full-resolution stills, along with the cached-still machinery. `activeStill` outlived it as a dangling field, which is what #103 was about. Nothing in the shell currently renders a WE project to an image at display resolution.

A *video* wallpaper can at least be sampled with ffmpeg. A **scene** project cannot — it has no decodable frames, only the renderer's output, which is precisely what was removed.

## Options, roughly in order of cost

1. **Sample video projects with ffmpeg at display resolution**, keep the preview for scenes. Partial, cheap, and covers the common case.
2. **Have the embedded WE surface capture one frame** once it exists (see the embed-wallpaperengine work) and publish it as the still. Correct for both project types, but blocked on that.
3. **Upscale the preview.** Rejected — invents detail and would still look soft.

Option 2 is the real fix and falls out of the embedded-renderer work almost for free; option 1 is worth doing before then only if scenes stay a minority of what people actually use.

## Not a regression

Before 0.14.9 the greeter showed a *months-old wallpaper from a different project* at full resolution. Now it shows the *correct* wallpaper at lower resolution. Strictly better, but not finished.

PR #118, opened 2026-08-05
fix(wallpapers): render the still offscreen, and only once per project

XephyLon:
Follow-up to #117. The still rendering worked, but the method was intrusive in two ways, both reported after using it:

- **it took over the screen** — `--window` opens an ordinary floating window, which takes focus and covers whatever was in front of it for the three seconds the render takes;
- **it did it every time** — including switching back to a scene seen a minute ago, re-rendering a file already on disk byte for byte.

## Offscreen

`--screen-root` binds a wlr-layer-shell surface instead of a window, and `--layer background` puts it on the lowest layer — underneath the shell's own wallpaper, which is opaque and covers it completely.

Verified on a live session mid-render:

```
windows named wallpaperengine: 0
active window:  unchanged
DP-1 level=0  linux-wallpaperengine     <- render
DP-1 level=1  quickshell:background     <- covers it
```

…and the frame still comes out at 5120x1440. That an occluded background layer keeps rendering at all was the part worth checking rather than assuming.

`--no-fullscreen-pause` is required alongside it: a background pauses by default while a fullscreen window is active, and a paused renderer never reaches the screenshot frame — a render started over a fullscreen app would sit until the timeout and produce nothing. In window mode the question never arose.

The window path stays as the fallback for anything without a named output: a non-Hyprland compositor, or an explicit geometry argument asking for a size that is not a monitor's.

## Cached

A project's render does not change, so it is kept and reused.

| | before | after |
|---|---|---|
| first apply | 3.20s | 3.20s |
| every apply after | 3.20s | **0.044s** |

The cached file's *size* is checked, not just its existence — a still cached at the old resolution is exactly the upscaled-thumbnail problem #113 was about, so a monitor change invalidates it. `WE_STILL_FORCE=1` re-renders regardless, for a project updated in the Workshop at unchanged dimensions.

The cache check sits ahead of the "is the renderer installed" guard deliberately: a usable still needs no renderer, and failing there would discard a good file on a machine that has since dropped the extra.

## Testing

Full suite green (276 passed, 0 failed). Each of the three commits passes its own tests, so the history bisects.

Mutation-tested — dropping `--layer background`, `--no-fullscreen-pause`, `--screen-root`, `--silent`, the size comparison, the `WE_STILL_FORCE` bypass, or the cache branch entirely each fail the suite.

One mutation survives and is equivalent rather than uncovered: swapping `-s` for `-e` on the cached file lets an empty file past the existence check, but `ffprobe` then reports no geometry, the comparison fails, and the cache misses anyway. `-s` is kept because it avoids spawning `ffprobe` on a zero-byte file.

Worth flagging about the tests themselves: the first version of the cache tests tried to simulate "no renderer" by narrowing `PATH`. `linux-wallpaperengine` lives in `/usr/bin`, so any `PATH` narrow enough to hide it also hides `bash` and `ffprobe` — those tests silently ran the *real* renderer and passed for the wrong reason. They now shadow it with a stub that records whether it ran, which is the actual question.

## Not addressed

Rendering a scene through a second process at all remains the wrong shape; the right fix is the shell owning the Wallpaper Engine texture, at which point a still is a framebuffer read and costs nothing. That is the parked embed work, unchanged by this.

### DIFF
--- a/dots/.config/quickshell/imi/scripts/wallpapers/we_still.sh
+++ b/dots/.config/quickshell/imi/scripts/wallpapers/we_still.sh
@@ -26,10 +26,16 @@ GEOMETRY="${3:-}"
 command -v linux-wallpaperengine >/dev/null || { echo "we_still.sh: linux-wallpaperengine not installed" >&2; exit 3; }
 
 # Default to the focused monitor: the greeter fills one screen, and rendering
-# at that size is what makes the result native rather than upscaled.
+# at that size is what makes the result native rather than upscaled. The name
+# is wanted as well as the size - see the render mode below.
+MONITOR_NAME=""
 if [[ -z "$GEOMETRY" ]]; then
-    GEOMETRY="$(hyprctl monitors -j 2>/dev/null \
-        | jq -r 'map(select(.focused)) | .[0] | "\(.width)x\(.height)"' 2>/dev/null || true)"
+    monitor_json="$(hyprctl monitors -j 2>/dev/null \
+        | jq -c 'map(select(.focused)) | .[0] // empty' 2>/dev/null || true)"
+    if [[ -n "$monitor_json" ]]; then
+        MONITOR_NAME="$(jq -r '.name // ""' <<<"$monitor_json" 2>/dev/null || true)"
+        GEOMETRY="$(jq -r '"\(.width)x\(.height)"' <<<"$monitor_json" 2>/dev/null || true)"
+    fi
 fi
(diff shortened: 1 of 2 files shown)


XephyLon:
Superseded by the branch that removes the subprocess entirely. Hiding the render window on a background layer and caching the result were both treatments of the symptom — the shell embeds Wallpaper Engine in-process, so the frame was always already on screen and never needed re-rendering.

PR #125, opened 2026-08-05
Proposal: greeter staging dir — root out of the hot path

XephyLon:
Design spec for Option B of the reactive greeter sync — taking root out of the greeter-update hot path via an sddm-readable staging directory. This is the separate proposal that `docs/superpowers/specs/2026-08-05-reactive-greeter-sync-design.md` promised; Option A (the `GreeterSync` observer + satellite sync wrapper with the fingerprint gate) is shipped and stays as-is until this is approved and implemented.

**Spec:** `docs/superpowers/specs/2026-08-05-greeter-staging-design.md`

## What it proposes

- **Install-time (root, once):** a staging directory at `/var/lib/imi-sddm-theme`, owner `user:sddm`, mode `2750` — setgid, because the user cannot `chgrp` files to a group they are not in, and the setgid bit is what makes unprivileged writes come out sddm-readable.
- **Theme resolution via four symlinks:** `Components/Settings.qml` and `Components/Colors.qml` *must* be symlinks (the registered singletons cannot be absolute-pathed through any conf); `Themes/ii-sddm.conf` and the `Backgrounds/` directory follow so there is one mechanism instead of two, and the conf keeps its relative paths.
- **Hot path:** the sync wrapper keeps its flock + fingerprint gate and replaces `sudo apply` with atomic user-space writes (tmp-in-same-dir + rename). Root executes nothing in the steady state; the NOPASSWD sudoers rule leaves the hot path (retire vs. keep-as-repair-lever is an open question).
- **Login safety:** the satellite's replacement-first / repoint-second / delete-last ordering applied to install, migration and uninstall; the critical pair (the QML singletons) only ever swapped by atomic rename, with the one non-atomic swap (the `Backgrounds` directory) confined to the link whose dangling is cosmetic. Optional tmpfiles.d + factory-copy hardening turns "staging wiped" from login-breaking into wrong-colored-but-working.
- **Security, stated honestly:** trust is a wash (root already couriers user-authored QML into the pre-auth greeter verbatim); what improves is no root ffmpeg/parsers on user-controlled inputs and no standing NOPASSWD grant; the single-user assumption becomes structural and the multi-user caveat is spelled out rather than papered over.
- **Migration:** all satellite (`setup.sh`/`uninstall.sh`/`check.sh`) + hub `SDDM_REF` bump; zero hub QML changes — `GreeterSync.qml` and its tests are untouched because A routed everything through the one wrapper.

Current state was verified against the real scripts in the satellite reference checkout (`setup.sh`, `uninstall.sh`, `check.sh`, `iiMatugen/sddm-theme-{apply,sync}.sh`, `generate_settings.py`) and the hub's `services/GreeterSync.qml`, not against the parent spec's summary.

## Open questions (answers wanted before implementation)

1. Ownership: plain `user:sddm 2750` (recommended) vs root-owned dir + ACLs?
2. Sudoers rule: retire outright, or keep a NOPASSWD root apply as a manual repair path?
3. Adopt the `factory/` + tmpfiles.d self-healing hardening?
4. Scope: `ii-matugen` mode only first, or all three install modes in one change?
5. Confirm symlinked `Backgrounds/` over absolute paths in the conf?
6. Also filter `generate_settings.py` to an allowlist of greeter-consumed keys (the flattened full config is world-readable today; staging already narrows it to the sddm user)?
7. Is `/var/lib/imi-sddm-theme` the right path?
8. Keep A's stamp-after-success discipline unchanged (recommended)?

Spec only — no implementation, no satellite edits, no behavior change anywhere.

Docs: not needed — proposal spec only, no behavior change

### DIFF
--- a/docs/superpowers/specs/2026-08-05-greeter-staging-design.md
+++ b/docs/superpowers/specs/2026-08-05-greeter-staging-design.md
@@ -1,0 +1,421 @@
+# Greeter staging dir — root out of the hot path (Option B)
+
+**Status:** proposal — spec only, no implementation
+**Repos:** immaterial-impulse (hub), imi-sddm-theme (satellite)
+**Parent:** `docs/superpowers/specs/2026-08-05-reactive-greeter-sync-design.md` (Option B
+section is the seed; Option A is shipped and working)
+
+## Problem
+
+Option A made the greeter sync reactive and gated: the hub's `GreeterSync` service
+observes the greeter-relevant config leaves, the satellite's `sddm-theme-sync.sh`
+fingerprints what the greeter consumes and escalates only on a real change. That fixed
+staleness and the still race — but every *real* change still ends in
+`sudo sddm-theme-apply.sh`: a root process, fired from a user-space observer chain, that
+parses user-controlled files and copies user-authored QML onto `/usr/share`. Root is
+still in the hot path; the NOPASSWD sudoers rule that lets it run unprompted is a
+standing grant the satellite's own installer has to defend with a path-ownership walk
+(`setup.sh setup_sudoers`) precisely because one mistake turns it into `NOPASSWD: ALL`.
+
+Option B moves the greeter's mutable inputs into an sddm-readable staging directory
+that the *user* owns. The hot path becomes atomic user-space writes; root's involvement
+shrinks to install and uninstall, which are already interactive `sudo` sessions.
+
+## Current state (A), verified against the real scripts
+
+Verified against the reference checkout at `/home/xephy/dev/imi-sddm-theme` and the hub
+tree — not against the parent spec's summary.
+
+**Hub — `services/GreeterSync.qml`.** A singleton observing
+`Config.options.wallpaperSelector.wallpaperEngine.{activeProject, activePath,
+activeType, activePreview, scaling}` and `Config.options.background.wallpaperPath`,
+debounced 1.5s, serialized (a request landing mid-run re-arms the debounce instead of
+stacking a second process). It runs exactly one thing:
+`~/.config/imi-sddm-theme/sddm-theme-sync.sh`, existence-checked at fire time — absent
+wrapper is a silent no-op, so the hub works on machines without the theme.
+`Background.captureGreeterStill` pokes `GreeterSync.request()` after a successful still
+write (`modules/imi/background/Background.qml:340`), closing the still race.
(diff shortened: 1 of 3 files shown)


XephyLon:
Implemented. The code lives in the satellite — **[imi-sddm-theme#15](https://github.com/XephyLon/imi-sddm-theme/pull/15)** — since that is where `setup.sh`, the apply scripts and `check.sh` are; this PR now carries the spec plus the `SDDM_REF` bump that makes it reach users. **Merge the satellite PR first**, or this pin names a commit on an unmerged branch.

## Decisions taken

Recommendations as written, with three notes:

- **Q8 → simplify**, as you asked. It cost nothing: with `set -e` and the stamp as the final statement, a failed apply still cannot be recorded as done, and the existing "failed apply is retried, not stamped" test passes unchanged. The statement ordering is now pinned so the simplification cannot rot.
- **Q4 was forced by Q2.** The sudoers rule cannot retire while matugen-only's `post_hook` still calls the root apply directly, so all three modes converted together rather than sequencing `ii-matugen` first.
- **Q6 deferred** as orthogonal. Staging already moves `Settings.qml` from world-readable under `/usr/share` to `0640` in a `2750` directory; allowlisting the flattened keys remains a separate improvement.

## What a reviewer should look hardest at

The **setgid bit** (the user cannot `chgrp` into the `sddm` group, so without it every published file is unreadable by the greeter — a staging dir that looks right and a login screen with no background), and the **link-severity asymmetry** that the install ordering depends on: the `Settings.qml`/`Colors.qml` pair swap with `rename(2)` because dangling means no graphical login, while `Backgrounds` — whose failure is only the fallback colour — absorbs the one swap that cannot be atomic.

## Verification

Nothing was installed and no privileged command ran at any point. Everything was driven against sandboxed trees through the `SDDM_*`, `IMI_SDDM_STAGING` and `IMI_SDDM_USER_HOME` seams, with `sudo` stubs refusing anything bound for `/etc`, `/usr` or `/var/lib`. **`/etc/sddm.conf` and every file in `/etc/sddm.conf.d/` are byte-identical to before this work**, and no `/var/lib/imi-sddm-theme` or tmpfiles snippet exists on my machine.

Both pin sites moved together and `test_sddm_theme_source.py` is green; full hub suite 337 passing.

Docs: not needed — hub-side change is the pin plus this spec's status; the design and its rules are documented in the satellite's AGENTS.md (§The greeter staging directory).

PR #128, opened 2026-08-06
fix(settings): cursor themes pick from a card grid, not a chip strip

XephyLon:
UX report: the cursor theme picker used `ConfigSelectionArray`, whose options render as a single horizontal chip row — with 15 installed themes it ran off the page edge and clipped its own label.

Replaced with the icon-pack selector's card-grid idiom: three columns, wraps to fit, active theme ringed + check icon, click to apply (same `CursorThemes.apply` path, size preserved). No previews by design — cursor themes ship Xcursor binaries Qt can't decode, so cards are name-only.

Verified live on the running shell (hot reload): all fifteen themes visible simultaneously, active highlight on `Bibata-Modern-Classic`, screenshot-confirmed. Suite green, `DesignSystemCompile checked=136 failures=0` (compiles the page).

Docs: not needed — widget swap on one settings page, no mechanism or schema change

### DIFF
--- a/dots/.config/quickshell/imi/modules/imi/settings/pages/CursorConfig.qml
+++ b/dots/.config/quickshell/imi/modules/imi/settings/pages/CursorConfig.qml
@@ -67,23 +67,70 @@ ContentPage {
                 color: Appearance.colors.colSubtext
             }
 
-            GroupedList {
-                ConfigSelectionArray {
-                    text: Translation.tr("Theme")
-                    icon: "mouse"
-                    visible: CursorThemes.available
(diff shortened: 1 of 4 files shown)


XephyLon:
Extended per request: cards now show each theme's **actual pointer**, extracted by the scanner from the theme's own Xcursor file (stdlib struct+zlib parser — the format is trivial even though Qt can't decode it; no new dependencies). 18/18 themes on this machine extract correctly, screenshot-verified live; unparseable pointers degrade to the icon fallback. Synthetic-fixture tests pin the parser, frame selection, and degradation.

PR #138, opened 2026-08-09
fix(install): pin the SDDM theme to v0.3.0's commit, not a pre-rebase head

XephyLon:
#125 pinned `SDDM_REF` to the satellite branch head I had tested. That repo merges with **rebase**, so the commit was rewritten on landing: the pinned SHA is not an ancestor of `imi-sddm-theme`'s `main` and survives only on the merged branch ref. It fetches fine today and 404s the day that branch is deleted.

Re-pinned to `3e961a7` — the **v0.3.0 release commit**, which is the established shape here: the pin before #125 was v0.2.4's commit.

## Mechanized, not commented

A comment would not have helped, since rebase-and-merge is the default and this will recur. `tests/test_sddm_pin_reachable.py` asks GitHub whether the pinned SHA is an ancestor of the satellite's `main` and fails when it is not:

| SHA | compare vs main | verdict |
|---|---|---|
| `bde1f15` (pre-rebase head) | `diverged` | fails |
| `3e961a7` (v0.3.0 commit) | `identical` | passes |

It skips cleanly without network, so the suite stays green offline. Pin *parity* between the two sites is still `test_sddm_theme_source.py`; this is about whether the SHA they agree on actually landed.

## A claim I repeated without checking

I wrote "this repo has no tags" into #125's body, a code comment and a test docstring. It is false — the satellite tags every release, `v0.2.0` onwards — and I took it from a stale line in its AGENTS.md instead of running `git tag -l`. The pin is a SHA because `test_sddm_theme_source.py` requires one so no moving ref can change the theme under a pinned release, not because there was nothing to tag. Corrected here and in the satellite's AGENTS.md and CHANGELOG ([3e961a7](https://github.com/XephyLon/imi-sddm-theme/commit/3e961a7)).

Worth noting for the doc rules: the `Docs:` receipt guarantees an author documents *their own change*. It cannot catch a pre-existing sentence elsewhere in the file that has become false — which is how this survived a PR that did update AGENTS.md.

Full suite green (337), including the new check.

Docs: updated satellite AGENTS.md §What this is and CHANGELOG (the tags claim); hub-side is the pin plus the new test, both self-documenting.

### DIFF
--- a/dots/.config/quickshell/imi/tests/run_tests.sh
+++ b/dots/.config/quickshell/imi/tests/run_tests.sh
@@ -237,6 +237,12 @@ if ! python3 "$SCRIPT_DIR/test_tmux_dots.py"; then
     exit 1
 fi
 
+echo "Running SDDM pin reachability test..."
+if ! python3 "$SCRIPT_DIR/test_sddm_pin_reachable.py"; then
+    echo "SDDM pin reachability test failed."
+    exit 1
+fi
+
 echo "Running parallax migration runtime tests..."
 if ! python3 "$SCRIPT_DIR/test_parallax_migration_runtime.py"; then
     echo "Parallax migration runtime tests failed."
--- a/dots/.config/quickshell/imi/tests/test_sddm_pin_reachable.py
+++ b/dots/.config/quickshell/imi/tests/test_sddm_pin_reachable.py
@@ -1,0 +1,89 @@
+#!/usr/bin/env python3
+"""The pinned SDDM_REF must be a commit on the satellite's main branch.
+
+The satellite tags its releases, but the hub pins the SHA of the release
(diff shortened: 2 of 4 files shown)


XephyLon:
Updated: the pin now names **v0.3.1** (`1b0ac8b`) rather than v0.3.0.

imi-sddm-theme#16 merged after this branch was opened, so v0.3.0's commit would have shipped the greeter staging work *without* the last-resort wallpaper path fix — the one that made a fresh matugen-only install fail its very first apply. One merge here now delivers both.

Both pin sites moved together; `test_sddm_theme_source.py` (parity) and `test_sddm_pin_reachable.py` (the SHA is an ancestor of the satellite's main) are green, as is the full hub suite at 337.

PR #160, opened 2026-08-10
feat(widgets): per-widget grid sizes and a resize handle

XephyLon:
Steps 1-3 of `docs/superpowers/specs/2026-08-10-media-widget-sizes-design.md`.

A widget's span comes from its manifest, which is a constant, so no placed widget can be resized.
This adds the machinery generically rather than inside one plugin: `notes`, `user-card` and
`image-converter` all declare grids and all had the same one-size limitation.

- `manifest.grid.sizes` — the spans a widget actually has layouts for. **Absent means one size and
  no handle**, which is every shipped manifest today, so nothing changes for any of them.
- `__gridSize` in plugin state, resolved stored → manifest default → content-sized. A stored span the
  manifest no longer offers falls back rather than being honoured.
- The resize grip, shown only when more than one span is offered, previewing live and committing on
  release, Escape cancelling.
- `lint_plugin_option_keys.py` reserves the `__` prefix for host state so a plugin cannot collide.

**Not every widget is resizable, by design.** A widget qualifies only when it has a design per size —
offering a span with no layout behind it is worse than offering nothing.

## Two corrections to the spec

- The spec said the grip must claim the press before drag-to-move, which is true, but the *reason*
  given was wrong: nesting alone claims it, because `AbstractWidget` computes its drag by hand and
  has no `drag` target to steal with. Proved by removing `preventStealing` and watching the runtime
  harness stay green. The flag is kept and the comment rewritten to name the real mechanism.
- The spec said pointer grabs are unreachable from tests. The repo already has a `qs -p` +
  headless-weston pattern that reaches them, so there is now a 22-check runtime harness — including
  *the widget does not move* on every gesture, and a single-span control whose corner must move it.

## Worth a decision

`__gridSize` lives in `pluginOptions`, which presets capture and replace — so a preset now carries
widget sizes as it already carries positions. Confirmed intended.

## Testing

`./tests/run_tests.sh` — **390 passed, 0 failed** (366 baseline; +22 `tst_grid_sizes`, +2 validator).
`DesignSystemCompile` 154 files. Non-vacuity proved on committed trees for all four `gridSizes.js`
mutations, the reserved-prefix lint, the seven grip source pins and the runtime harness — one early
test *was* vacuous under a mutation and was rewritten until it reddened.

Docs: updated AGENT.md §Dynamic/data-driven QML gotchas, docs/widget-grid.md, CONTRIBUTING.md

### DIFF
--- a/AGENT.md
+++ b/AGENT.md
@@ -1027,6 +1027,48 @@ arrays, etc.) rather than static declarations - e.g. the plugin system in
   values. Desktop backdrop blur is also per-plugin state: a manifest opts into its default with
   `desktopWidget.blur: true`, while the generated **Blur background** option always lets the user
   override it. Do not make `PluginWidget` blur every plugin unconditionally.
+- **A manifest's option keys and the host's own per-plugin state are one namespace, so `__` is
+  reserved.** Both land in `pluginOptions[pluginId][key]` in `plugin-state.json`: the manifest's
+  declared options *and* the host's `blurEnabled`/`positionLocked`/`clickThrough`/`keepTranslucent`
+  seeds, plus `__gridSize` (the span a user resized a grid widget to). A manifest declaring a
+  `__`-prefixed key would therefore ship a settings control in Settings > Widgets that writes
+  straight over host state. `PluginValidator.js` rejects it - which is the only defence against an
+  installed third-party plugin, and silent for a bundled one, since a rejected manifest simply does
+  not appear - so `tests/lint_plugin_option_keys.py` is the half that names the file. Anything the
+  host stores per plugin from now on goes under that prefix; anything a manifest may set does not.
+  66c022fbe ("feat(plugins): reserve the `__` option-key prefix for host state").
+- **A widget's grid span is resolved, not read.** `manifest.grid` may offer several spans
+  (`grid.sizes`) with `cols`/`rows` as the default, and the resolution - stored choice, then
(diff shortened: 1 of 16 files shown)


XephyLon:
## Correction: this was inert in production until now

A follow-on branch found that `offeredSizes` gated its whole `sizes` list on `Array.isArray`, **which is not the test**. A JS array reaching a delegate through a `Repeater`'s `model` has crossed into QVariant and back: its indices and `length` survive, `Array.isArray` does not — and `Background.qml` builds every desktop widget from exactly such a model.

So a manifest declaring three spans arrived at the host offering one: **no grip, no resizing, no error anywhere.** Everything in this PR worked in its tests and would have done nothing on a real desktop.

Verified independently with an offscreen probe before accepting the claim:

```qml
Repeater {
    model: [{ grid: { sizes: [{cols:3,rows:2},{cols:2,rows:2}] } }]
    delegate: Item {
        required property var modelData
        Component.onCompleted: /* Array.isArray(modelData.grid.sizes) === false */
    }
}
```

Fixed here as `677d6ff05` rather than downstream, so this PR is correct on its own. `asSizeList()` accepts anything with a numeric `length`, still rejecting a malformed `sizes` rather than reading it as an empty list.

**Why the harness missed it:** it declares its manifests inline on the harness root, so they never cross a model boundary. It now also builds one widget through a `Repeater` and scores that the spans arrive — which is the check that would have caught this.

**393 passed, 0 failed** (was 390; +3 from the fix's own cases).

PR #161, opened 2026-08-10
feat(widgets): a cookie whose lobes move independently

XephyLon:
Step 4-5 of `docs/superpowers/specs/2026-08-10-media-widget-sizes-design.md`. Depends on #159 for a
live band source.

`MaterialCookie` builds `RoundedPolygon.star(sides, radius, innerRadius, rounding)` where
`innerRadius` is **one scalar for the whole shape**. Every lobe is identical, so animating it makes
the cookie breathe and can never make it a per-band visualizer.

- `RoundedPolygon.starPerLobe()` — an array of inner radii, one per lobe, falling back to scalar
  behaviour when handed a number. `star()` is untouched; the Material shape catalogue depends on it.
- `VisualizerCookie.qml` — `lobes`, `levels`, `baseRadius`, `reach`.
- `visualizer_bands.js` — folds the producer's bands onto the lobes in contiguous groups (bass at
  lobe 0, treble at the last) with a per-lobe attack/decay envelope, because raw values jitter at
  frame rate and make the outline boil.

## Measured, not assumed

The spec left performance open, with a fixed-rate-plus-interpolation fallback if per-frame
regeneration proved expensive. Offscreen: polygon rebuild **0.59 ms**, 0.77 ms per cookie per
animated frame at 240px, and **four cookies at once held 62 fps**. The fallback is not needed. A
frame where no lobe moved neither reassigns `levels` nor rebuilds, and the envelope tail snaps so the
shape reaches exact rest instead of decaying forever.

Also found: `ShapeCanvas` cannot host this — it builds a `Morph` and starts a 350 ms transition on
*every* polygon change (1.33 ms construction). The component draws its cubics onto a plain `Canvas`.

## Inner radii, not tips

The spec's wording read like the lobe tips move. Both were rendered offscreen before choosing: inner
radii keep the tips fixed, so the shape's bounds do not jump and a refit cannot rescale the whole
cookie every frame. `reach` defaults to 0.14 — at 0.18 a frame where several bands peak together is
indistinguishable from a circle.

## Testing

`./tests/run_tests.sh` — **393 passed, 0 failed** (366 baseline; +11 lobe radii, +16 band folding).
Non-vacuity proved on a committed tree: forcing `radii[0]` reddens 6 of 11, dropping the group offset
reddens 6 of 16, removing the short-array fallback 1, collapsing attack into decay 1.

Shapes and motion are not testable — `qmltestrunner` cannot construct Quickshell types and the
software scene graph draws no `Canvas`.

Docs: updated AGENT.md §Dynamic/data-driven QML gotchas, docs/tests-README.md


XephyLon:
Rebased onto `fix/cava-service-producer` (#159) and retargeted at it, because the two branches turned out to disagree in a way only a combined tree showed.

This branch was written before `CavaService` had a producer, so it counted itself into the refcount by hand — the exact bookkeeping #159's `CavaRef` exists to replace, and which its new `test_cava_contract.py` forbids. Merging both cleanly gave a red suite. The claim is now declared, bound to the same condition the hand-rolled version tracked.

The upside: with a real producer behind `CavaService.values`, the cookie is **live** rather than resting. My spec sent the original agent at a dead API, so "the cookie will not move on screen" no longer applies.

**412 passed, 0 failed** on the stacked branch (385 from #159, +27 here). Merge #159 first.

PR #204, opened 2026-08-15
Privacy indicator: a control panel behind the pill

XephyLon:
The privacy popup could only report. It can now act, in a card that grows
out of the same surface rather than replacing it.

**Hover reads, click acts.** Hovering keeps the old summary. Clicking the pill
pins the card and opens it into the controls — `StyledPopup` already had
`pinnedOpen`/`dismissRequested`, and the overlay's focus grab already counts
the bar window as inside, so the click that pins cannot also dismiss it.

**Three levers, ordered by what they cost the target app.**
- *Mute* a mic stream (`pactl set-source-output-mute`) — scoped to one stream,
  reversible from the row that set it. A muted stream stays listed: the app
  still holds the microphone and can unmute itself.
- *Stop / pause / save* the shell's own recording and replay buffer.
- *Force stop* — destroys the app's PipeWire node (`pw-cli destroy`). The app
  keeps running but loses a stream it never offered to give up, so it is off
  unless Settings turns it on, and `CaptureControl` refuses it when off rather
  than trusting the UI to hide the button.
- *Revoke* a portal permission (`PermissionStore.DeletePermission`).

**Where it says no.** Silence would read as an assurance, so the panel states
its limits: a camera holder is a process with `/dev/video` open, not a stream
that can be handed back, so its row is marked rather than given a button that
kills the app; a cast owned by another app must be stopped from that app; and
the permission list explains that apps opening the device directly never appear
there at all — which on a machine with no sandboxed apps means it is empty, and
says so.

**Untrusted input.** Every argv comes from `capture_control.js`, which validates
ids and returns null for anything it will not build. The tests that matter most
are the refusals — `"52; rm -rf ~"` is a string beginning with a valid stream
index and must not become one.

**Verified live**, end to end, against a virtual source built for the purpose
(`module-null-sink` + `module-virtual-source`, so no physical microphone was
opened; both modules and the probe process were removed afterwards and the
removal confirmed):
- hover shows the summary card; clicking pins and grows it
- a recording stream appears as its own row with a mute button
- clicking mute → `pactl` reports `mute = True`, the row dims, the button reads
  toggled; clicking again → `mute = False`
- replay row shows save + stop; permission section shows the honest empty state
- a click outside dismisses the panel

Suite 719 passing (704 + 15 new).

**Not exercised by me:** force stop (`pw-cli destroy`) and revoke were tested as
command construction and refusal only — firing them needs a real sandboxed app
holding a portal grant, and a capture worth destroying.

Docs: not needed — no doc describes the privacy indicator's behaviour; the
non-obvious parts (why `object.id` is not the pactl index, why a muted stream
stays listed, what revoke cannot reach) are recorded in `capture_control.js` and
`CaptureControl.qml` where the next reader of those files will hit them.


XephyLon:
Pushed two more commits after the transition review.

**The transition was three faults, only one of them in the panel.**

1. *Two trees.* Each section was declared twice — once for the summary, once for
   the controls — with `visible` toggled between the copies, so everything the
   two depths share was destroyed and rebuilt: the header and section icons
   blinked out while the buttons popped in at full size. Now one tree. Every
   element present at both depths is declared once and stays on screen; only the
   buttons and the permission block arrive, growing from zero width the way the
   pill's own icon slots do. No fades on anything shared.

2. *The card was chasing its own content.* It eases width/height toward whatever
   it shows, so content that eases too is a target that keeps moving — the card
   trails the whole way, and since content is centred, the overhang is clipped
   (the header, off the top, for the length of the transition). `contentDrivesSize`
   on `StyledPopup` hands the card to the content for the duration: card takes the
   content's size and position directly, one animation, card exactly as big as
   what it holds every frame.

3. *A timer that could never fire.* The overlay retargets the card from a
   zero-interval timer, which coalesces bursts of size changes — but a popup
   animating its own size changes it every frame, so the timer was re-armed
   before it could fire and the card sat at its old size until the animation
   ended. Content-driven popups now retarget on the spot; everything else keeps
   the coalescing it was written for.

Verified frame-by-frame at ~50ms: every element present before the click is
present throughout, nothing clipped. Two frames still showed a cropped header —
that turned out to be the measurement: `grim` is a screen capture, so the
screenshot loop trips the screencast detector and adds a section mid-transition.
Confirmed by capturing the settled card with *no interaction at all* and seeing
the same artifact.

**Pointer shape.** Seven bar widgets acted on clicks while leaving the arrow
alone (privacy, tray items, media, weather, system icons, timer pill, util
buttons). The shared button types always set it; every hand-written MouseArea
forgot separately — so it is now `tests/lint_clickable_cursor.py`, which found
an eighth case while I was writing it. Hover-only areas stay arrows: `Resource`
takes `Qt.NoButton`, `SysTrayMenu` listens for back/right only, and a hand over
something that ignores a left click would be its own lie. Proven to fail on a
planted mutation, wired into `run_tests.sh`.

Suite 719 passing. The virtual-source probe used for testing was removed and the
removal confirmed (`0` modules left, no `parecord` running).

PR #205, opened 2026-08-15
Recording region: stop, pause and save-clip beside the capture

XephyLon:
A region recording gives no sign of itself inside the frame — which is the point
— but that also made stopping one a trip to the bar for a control that could be
an inch from where you were already looking. This puts stop, pause and (while
the replay buffer is armed) save-clip against the rectangle being captured, for
as long as the recording runs.

**The toolbar is never inside the region.** gsr captures whatever the compositor
draws in that rectangle, so a toolbar over it is a toolbar in every frame of a
clip that cannot be re-taken. Placement goes below the region first, above when
there is no room below, and — when neither fits, i.e. a full-height or
full-screen capture — places nothing at all rather than place it inside. The bar
indicator still stops the recording in that case, so what is lost is a shortcut,
not the ability. The test sweeps a grid of sizes and positions asserting the two
rects never intersect; planting an overlapping placement fails four tests.

**One writer for the recording state.** The first version wrote the region from
QML right after launching `record.sh`. The shell rewrites that state file
wholesale from its own in-memory copy, so the write landed on top of the
`enable = true` the script had made milliseconds earlier and reverted it — the
recording ran with the shell believing nothing was recording. `record.sh` already
knows the region, so it now writes the flag and the region in one `jq` update and
the round trip is gone. A full-screen capture stores an empty region.

**Verified live through the selector**, not by calling the script:
- dragged out a region, recording started (`enable: true`, region `1,117 600x360`,
  gsr running with `-w region -region 600x360+1+117`)
- toolbar appeared below the region, centred and clear of it
- clicked **pause** → `paused: true`, indicator switched to `pause_circle`, the
  button became play
- clicked **stop** → recording ended, state cleared by the script, 3.9 MB / 40 s
  file written
- extracted a frame from the last two seconds of that clip, while the toolbar was
  on screen: only wallpaper, no toolbar

Suite 730 passing. One pre-existing flake seen once and not reproducible
(`test_present_settings_controls_reach_the_daemon`, a clight runtime test —
passes standalone, unrelated to these files).

Test recordings made during verification were deleted afterwards.

Docs: not needed — the constraint that makes this design what it is (a region
capture films the compositor, so controls must live outside the rectangle) is
recorded at the top of `recording_region.js`, next to the placement rule it
dictates, and the single-writer reasoning is in `record.sh` above the function
it applies to.

### DIFF
--- a/dots/.config/hypr/hyprland/rules.lua
+++ b/dots/.config/hypr/hyprland/rules.lua
@@ -161,6 +161,13 @@ hl.layer_rule({ match = { namespace = "quickshell:popup" }, xray = false}) -- No
 hl.layer_rule({ match = { namespace = "quickshell:popup" }, ignore_alpha = 1}) -- No weird color for bar tooltips (but somehow this is necessary)
 hl.layer_rule({ match = { namespace = "quickshell:mediaControls" }, ignore_alpha = 1}) -- Same as above
 hl.layer_rule({ match = { namespace = "quickshell:reloadPopup" }, animation = "slide"})
+-- The recording controls are a small toolbar in a window that is mostly
+-- transparent: the extra room is for the toolbar's own shadow. The
+-- catch-all above blurs anything over 5% alpha, which frosts that shadow
+-- into a haze the size of the window. Same treatment as the popup and
+-- overlay surfaces, and for the same reason.
+hl.layer_rule({ match = { namespace = "quickshell:recordingRegion" }, ignore_alpha = 1})
+hl.layer_rule({ match = { namespace = "quickshell:recordingRegion" }, no_anim = true})
 hl.layer_rule({ match = { namespace = "quickshell:regionSelector" }, no_anim = true})
 hl.layer_rule({ match = { namespace = "quickshell:screenshot" }, no_anim = true})
 hl.layer_rule({ match = { namespace = "quickshell:session" }, blur = true})
--- a/dots/.config/quickshell/imi/modules/common/Persistent.qml
+++ b/dots/.config/quickshell/imi/modules/common/Persistent.qml
@@ -89,6 +89,12 @@ Singleton {
 
             property JsonObject record: JsonObject {
                 property bool enable: false
+                // The geometry a region recording was started with, "X,Y WxH",
+                // or empty for a full-screen one. Persisted beside `enable`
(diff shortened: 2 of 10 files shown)


XephyLon:
Confirmed working on the live shell. Final state of the branch:

**Look.** Rebuilt on the shell's own M3 expressive toolbar components — `Toolbar`,
`IconToolbarButton`, `ToolbarPairedFab` — so it matches the region selector's
toolbar rather than being a hand-rolled pill. Stop is a paired FAB in the error
role; the status icon breathes while recording and holds still when paused. The
FAB sits at a wider step than the toolbar's internal rhythm so it reads as paired
with the toolbar, not as a fourth button in it.

**Three bugs found by using it, none of them in the placement math:**

1. *Frosted shadow.* `rules.lua` blurs every `quickshell:.*` layer above 5% alpha,
   which turned the toolbar's shadow into a haze the size of its window. Added the
   `ignore_alpha = 1` rule that `popup` and `mediaControls` already carry.

2. *Dead click areas and a stuck pointer.* `StyledToolTip` is a Controls Popup,
   and a Popup renders in the overlay of its own window — here, a window the exact
   size of the toolbar. The tooltip had nowhere to go but on top of the buttons,
   arrived with no delay, and while up it owned the pointer: arrow cursor, clicks
   swallowed. Removed, with the reason recorded in the file so they don't come
   back without a surface of their own.

3. *`RippleButton`'s interactive area.* It declares its MouseArea in the body of a
   Control, so Qt parents it into `contentItem`. With the default contentItem that
   covers nearly the whole button and nobody noticed across 101 files — but any
   subclass replacing the contentItem collapses the area onto it. Fixed by
   parenting the MouseArea to the button and adding a `HoverHandler` for the
   pointer shape (handlers aren't Items, so `contentData` rules can't narrow them).

**This PR touches shared things**, worth knowing before merge:
- `modules/common/widgets/RippleButton.qml` — interactive area, all buttons
- `modules/common/widgets/ToolbarPairedFab.qml` — four colour aliases, defaults unchanged
- `dots/.config/hypr/hyprland/rules.lua` — one new layer rule
- `scripts/videos/record.sh` — now the single writer of the `record.*` state block

Suite 730 passing.

PR #295, opened 2026-08-24
Settings honesty and one bounded decode: EasyEffects, Screens, selector backdrop

XephyLon:
Three items from the task list, plus one environmental test repair found on the way:

- **fix(easyeffects):** the quick toggle flipped `active` optimistically and nothing ever re-checked, so a failed launch left it saying on for the rest of the session. It still answers the click at once, and verifies against the real process list one grace period later. New weston harness (`EasyEffectsStateRuntimeTest`) drives all three outcomes against stateful fake binaries; proven red on the planted original bug.
- **fix(settings):** the Screens section hid itself on one monitor while staying in the search index — the Clight section's bug (f9311591) one page over, plus a sharper edge: a laptop undocked with a stored screen filter hid the only control that can clear it. The section stays; the chooser shows when it can mean something; one screen with no filter gets a caption. Verified on this machine's live shell (single monitor) — search lands on the section.
- **perf(wallpapers):** the selector's blurred backdrop decoded the wallpaper at file resolution on every open; now bounded to the item it draws into. The sweep found no other unbounded wallpaper decode; the frost's shared-request caveat (33139b688) is why Background's own request is untouched.
- **test(clight):** the not-installed harness flavour faked absence by omitting a fake binary, which broke the day clight got genuinely installed on this machine — an sh shim now answers the one detection probe.

Suite green end to end (1186 QML checks, DesignSystemCompile 185 files, all runtime harnesses including the new one).

Docs: not needed — no mechanism changed; the rowVisible/section pattern is already documented in AGENT.md from f9311591's series
Changelog: not user-visible — entries ride the next catch-all release PR, per the release workflow

### DIFF
--- a/dots/.config/quickshell/imi/EasyEffectsStateRuntimeTest.qml
+++ b/dots/.config/quickshell/imi/EasyEffectsStateRuntimeTest.qml
@@ -1,0 +1,112 @@
+import QtQuick
+import Quickshell
+import qs.services
+
+/**
+ * EasyEffects toggle honesty, against the real singleton and fake binaries.
+ *
+ * Driven by tests/test_easyeffects_state_runtime.py, which puts fake
+ * `easyeffects`/`flatpak`/`pidof`/`pkill` executables at the front of PATH.
+ * The fakes keep a little state of their own in $EE_STATE_DIR: launching
+ * succeeds only while `launch_ok` exists (and then leaves a `running` marker
+ * pidof answers from; pkill removes it), so the harness can drive both a
+ * launch that fails and one that works.
+ *
+ * What it pins: enable() answers the click optimistically, and the verify
+ * pass corrects the lie - a failed launch reads as off again within
+ * verifyInterval, a real one stays on, and a kill stays off.
+ *
+ *   EE_STATE_DIR=... qs -p EasyEffectsStateRuntimeTest.qml
+ */
+ShellRoot {
+    id: harness
+
+    property int failures: 0
+    property int checksRun: 0
+    property int step: 0
+    readonly property string stateDir: Quickshell.env("EE_STATE_DIR") ?? ""
+
+    function check(label, ok) {
+        harness.checksRun++;
+        console.log(`[EasyEffectsState] ${label}: ${ok ? "ok" : "FAIL"}`);
+        if (!ok)
+            harness.failures++;
+    }
+
+    function finish() {
+        console.log(`[EasyEffectsState] checks: ${harness.checksRun} failures: ${harness.failures}`);
+        Qt.exit(harness.failures === 0 ? 0 : 1);
+    }
+
+    Component.onCompleted: {
+        // The suite cannot wait out the production 1.5s grace.
+        EasyEffects.verifyInterval = 300;
+    }
+
+    Timer {
+        id: waitForDetection
+        interval: 200
+        repeat: true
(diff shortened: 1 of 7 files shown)


XephyLon (commented):
Reviewed all four commits. EasyEffects verify design, the harness with stateful fakes, the Screens section repair and the sh shim are all sound; CI green. Two findings, one measured.

**1. The backdrop `sourceSize` binding decodes at full resolution first — the fix as written makes open #1 worse, not better.**

`sourceSize.width: width` on an `anchors.fill` item reads 0 at the moment the load starts (anchors resolve after creation), and Qt treats a 0 sourceSize as unbounded. Measured with a qml6 probe reproducing the exact shape (anchored Image, geometry resolving late, 3840x2160 source):

```
LOADING sourceSize=0x0
READY   decodedAt=3840x2160     <- the unbounded decode the commit set out to remove
LOADING sourceSize=300x0        <- each axis binding then kicks its own reload
LOADING sourceSize=300x169
READY   decodedAt=300x169
```

So the first open pays the full-res decode AND two re-decodes. Any later geometry change (searchbar toggle resizing the grid area) re-decodes again, since sourceSize changes force a reload.

Fix: bind to the constants the selector already sizes itself from, known at creation and stable for the life of the window:

```qml
sourceSize.width: Appearance.sizes.wallpaperSelectorWidth
sourceSize.height: Appearance.sizes.wallpaperSelectorHeight
```

Slightly larger than the item (the item sits inside the card's margins), which PreserveAspectCrop absorbs — one bounded decode, zero re-decodes.

**2. A stale verify result can briefly overrule a fresh click (low, self-correcting).**

Sequence: enable() at t=0; verifyTimer fires at 1.5s and fetchActiveStateProc starts; the user clicks disable() at 1.55s (active=false, timer restarted); the in-flight probe exits at ~1.6s, still sees the daemon (pkill hasn't landed) and writes `active = true` — the toggle shows ON for up to one grace period after an OFF click, until disable's own verify corrects it. Same shape for the startup `running: true` probe racing an early click.

One-line fix in the probe's onExited: results are only trustworthy when no newer toggle is pending —

```qml
onExited: (exitCode, exitStatus) => {
    if (verifyTimer.running) return; // a newer toggle owns the state
    root.active = exitCode === 0
}
```

The harness can pin it by clicking disable between a verify's start and its exit, though the timing window makes that fiddly to drive deterministically; the guard is cheap either way.

XephyLon (commented):
LGTM - both findings addressed the right way (GitHub refuses a formal approve on one's own PR, so this comment is the verdict). The backdrop bound is the creation-time constants with the measurement recorded at the call site, and the verify race is closed from both ends (discard-while-pending + restart-not-join) with the harness driving the stale landing deterministically via the slow-mode pidof - the timeline checks out (probe answers ~T700 against a T950 verify), and EXPECTED_CHECKS moved with the step list.

One residual micro-edge, non-blocking: the restart path (`running = false` on a live probe) delivers that probe's onExited with a kill exit code; if the timer has already stopped by then, the guard admits it and writes a spurious `active = false` blip until the fresh probe corrects it - reachable only when a verify restarts a still-in-flight probe, self-correcting within one probe round-trip. `if (exitStatus !== Process.NormalExit) return` in onExited closes it whenever it next annoys someone; not worth holding the merge for.

PR #296, opened 2026-08-24
The motion language's missing half: container-then-fill, one entrance spelling, text and list motion — plus the AI upgrade proposal

XephyLon:
Five streams, integrated and verified together (full suite: 1218 QML checks, all harnesses, DesignSystemCompile 185+).

**Bar popups arrive container-then-fill.** The card's below-the-fold sections hold until `card.openProgress` clears `Appearance.animation.contentsArrived` (no leadIn — the card is the second surface with a real progress to gate on), then cascade by visible rank. The hero section deliberately stays undressed: the card opens at its height precisely so it is legible on frame one. Exits ride out rigid. One wave on the overlay, never per popup; the wave arms only on a fresh open so takeovers and re-hovered exits never blink content. `lint_bar_popup_overlay_static` grew a gate section; a per-popup register (`test_bar_popup_section_entrance`) pins hero/channel/opt-in shape; nine planted mutations all caught.

**The three-channel entrance is one spelling.** `StaggerEntrance` (a QtObject beside the wave) installs opacity, a derived near-1 scale and a small rise on every member declaring `appear` — the drawer's nine hand-spelled dressings collapse into it with its exact values pinned, RippleButtons are skipped at runtime (their control owns the fold), and both composition lints learned to see the new spelling (one real detection gap found and fixed by the planted proof). Desktop menu rows adopted; **dialogs refused** into `STAGGER_DECLINED` with tree-read arguments (a cascade on a modal is latency on the question the user was interrupted to read; polkit focuses a field that would still be parked invisible).

**Text and lists move.** Media label and weather temperature slide on change (`animateChange`, already the house mechanism); clock digits refused — a seconds-capable format would tick an animation on a Top-layer bar that stays `visible` under fullscreen (53d1ff893's occlusion). Todo lists ride `StyledListView`'s own add/remove transitions, which required a real fix: the per-update `Object.assign` wrapper broke ScriptModel's identity diff, so every change had been remove-all+add-all — transitions could never fire, and stale indices across deletes are gone with it.

**AI: proposal + stage 1.** `docs/proposals/ai-assistant-upgrade.md` maps the surveyed fork's provider catalog / per-tool permissions / reviewed-mutation shapes onto our stack in five stages with eight open questions; stage 1 lands: `services/ai_catalog.js`, pure and testable, contract-pinned bidirectionally against `Ai.qml`'s literals so neither drifts before stage 2 wires them. Two defects in our tree recorded as evidence: `set_shell_config` writes config unapproved and never answers the model (the turn stalls), and custom-provider keyring keys are index-keyed so removals shift later providers off their keys.

**The sidebars materialize composed — the fork's real motion language, re-read object by object.** The right sidebar's section wave came back the register's way (the two refusal facts had been repaired by unrelated work), and the first re-adoption — gated on the slide's progress through `contentsArrived` — was measured on screen as ruined: parked content plus a gate puts the construction on stage. What shipped is the opposite: the wave runs ungated on the open's rising edge, *under* the slide, whose new `EdgeSlide.reveal` curtain (≤15% panel opacity for the first 40% of travel) masks the early frames so the panel arrives already composed and only the last-ranked members visibly land. On top of that: quick-toggle tiles converge from their own side of the grid (`convergeFrom`/`convergeSettle` in the policy), slider fills sweep from zero with the icon turning with the fill, the calendar ripples diagonally, the notification status row converges, and the left sidebar and AI pane take the same cascade. Along the way the tiles' leftover `Behavior on opacity` that swallowed the wave was removed and a lint now fails any wave member carrying one; a member hidden at the open is settled rather than left parked; a tile added mid-session fades in; the entrance survives `keepRightSidebarLoaded` being off; and a `wavePending` armed for one popup tree no longer fires against its replacement.

Exit tiers untouched: 2:1 stays per the maintainer's call, 4:1 A/B remains one central change if ever wanted.

Docs: updated AGENT.md §Design language
Changelog: updated — two Changed entries and one Fixed under [Unreleased]

### DIFF
--- a/AGENT.md
+++ b/AGENT.md
@@ -4478,6 +4478,43 @@ arrived.
 ("feat(widgets): StaggerWave, the one runner for a group's arrival";
 "feat(editMode): the drawer arrives as a surface and then fills").
 
+**What a member's ARRIVAL looks like is one spelling too — `StaggerEntrance`, beside the wave.**
+The three-channel entrance (opacity, a scale from a derived near-1 start, and a small rise, all on
+the one `appear` scalar the wave animates) lived only as Edit Mode's drawer's local dressing,
+spelled out per member nine times — the template the next adopter would copy, and a hand-copied
+dressing is how a member arrives on two channels out of three. `modules/common/widgets/
+StaggerEntrance.qml` is the promotion: declared beside the `StaggerWave`, aimed at the same
+container, it installs the three channels once on every child declaring `appear` — a member's whole
+opt-in is the property the runner already requires. The scale's START is derived, not picked:
+`motion_policy.js`'s `entranceScaleFrom(rise, reference)` matches the scale's excursion to the rise
+(the survey's raw 0.85 is a popup number that reads as a zoom on a full-width row) with the
+measured 0.85 as the floor, and `Appearance.animation.entranceRise` is the rise
+(`spacing.space250` — a distance on the shell's rhythm, not a duration). Two refusals are
+load-bearing: a child without `appear` is not a member, and a child owning an `interactionMotion`
+is skipped whole — a `RippleButton` already folds `appear` into the opacity binding that carries
+its disabled dim and owns `scale` through the model, so a writer here would REPLACE those bindings
+rather than compose with them. The runtime skip cannot see an ANCESTOR, so both composition lints
+learned the spelling: `lint_interaction_motion_double.py` fails a `StaggerEntrance` inside a
+scale-channel control and `lint_disabled_opacity.py` fails one anywhere in a file whose root dims —
+at ANY indent, because the realistic placement is a direct child at one level, which the ≥5-space
+nested-dim convention never sees (the plant proving the rule landed exactly there and the first
+version stayed green over it). A wave whose members are not one container's children — GroupedList
+reparents each row into its own plate — hands the runner the declared list via `StaggerWave.items`,
+which REPLACES the children walk. The desktop menu's rows adopted the entrance on the drawer's
+argument (its card's opacity IS a QML-readable container progress, so the wave gates with no
+`leadIn`; enter-only, since the close destroys the window); dialog content was assessed beside it
+and REFUSED — a dialog's content is the question the user was interrupted to read, and the polkit
+prompt focuses a password field the cascade would leave parked invisible under real keystrokes —
+with the full argument in `STAGGER_DECLINED`'s entry.
+("feat(motion): a wave member's entrance scale is derived, in the policy",
+"feat(widgets): StaggerEntrance, one spelling of a wave member's arrival",
+"refactor(editMode): the drawer's entrance dressing becomes the shared spelling",
+"test(lint): the two composition lints see the shared entrance dressing",
+"fix(lint): a dresser one indent deep is inside the control too",
+"feat(widgets): a StaggerWave can take its members as a list",
+"feat(desktopMenu): the menu's rows arrive in sequence, gated on the card",
+"test(motion): dialogs decline the group entrance, in the register".)
+
 **A wave asked for while its container is off screen writes nothing, and leaves the surface blank
 for ever.** Ranking asks each member whether it is on screen, and `visible` is EFFECTIVE visibility
 (see the `GroupedList` entry under
--- a/AGENT.md
+++ b/AGENT.md
@@ -4531,6 +4531,25 @@ A surface whose trigger is already the container's own `visible` (`ContentPage`,
 settings window's page switch) never showed it.
 ("fix(widgets): a wave waits for its container to be on screen").
 
+**A wave member must not carry a `Behavior on opacity`, because the member's opacity IS the wave's
+channel.** The wave assigns `appear = 0` to park a member — a snap, deliberately — and animates
+`appear` to bring it in, with opacity riding that scalar through a binding. A root-level
+`Behavior on opacity` in the member's file intercepts every write that binding makes: the park
+becomes a 200ms **on-stage fade-out** (measured live: `appear=0` with drawn opacity still 1.000 at
+the open), and the wave's own per-frame animation retargets the Behavior every frame — b710ef731's
+frozen-Behavior shape — pinning drawn opacity at 0.147 while `appear` was already at 0.955, so the
+member lands a whole fade-length after its slot. The android quick toggles shipped exactly that:
+their old `opacity: 0` + `onCompleted` self-fade was retired in favour of the wave and the Behavior
+that had animated it was left behind, which on every right-sidebar open drew the first-ranked tiles
+fully lit as the panel edge appeared, dimmed them on stage, held a blank beat as the slide landed,
+and then ran the cascade — read by the maintainer as "three quick toggles visible at all times,
+then the animation begins", through several attempted fixes aimed elsewhere.
+`tests/lint_wave_member_opacity_behavior.py` fails the suite on a root-level `appear`/`Behavior on
+opacity` pair; nested Behaviors (a ripple overlay, a scroll shadow) animate a child's own opacity
+and stay free.
+15688f7c ("fix(quickToggles): drop the tile's leftover opacity Behavior that swallowed the wave"),
+a5ef0c29 ("test(lint): fail on a wave member carrying a Behavior on opacity").
+
 **The adoption is the thing that decays, so the adoption is what is pinned.**
 `docs/p3drovfx-motion-measured-2026-08-22.md` §4.2 measured the sibling fork's motion off screen and
 found our arithmetic correct, our guideline written, and the wiring at **three** files against their
--- a/AGENT.md
+++ b/AGENT.md
@@ -4574,11 +4574,19 @@ the moment it opens and the moment it closes."* It is off that surface again, an
 **It since came back, the register's way — argued, at the maintainer's request, after both
 refusal facts had been repaired by unrelated work**: the persistent EdgeSlide surface removed the
 per-gesture teardown that was the measured frame drop, and the slide's in-client `progress`
-scalar is exactly the container progress whose absence forced the guessed leadIn. The re-adopted
-wave gates on it through `contentsArrived`, carries no leadIn, and is enter-only (sections ride
-the slide out rigid). The register entry moved from `STAGGER_DECLINED` to `STAGGER_ADOPTERS` with
+scalar gave the wave something real to coordinate with. The first re-adoption gated on that
+scalar through `contentsArrived` — correct-sounding, and measured on screen as ruined: parked
+content plus a gate puts the whole construction ON STAGE, the panel arriving empty and visibly
+building itself. The design that stuck is the opposite, read object-by-object off the fork's
+re-recorded sidebars: the wave runs UNGATED, park-and-enter on the open's rising edge, UNDER the
+slide — whose `reveal` curtain (EdgeSlide's 15%-plateau opacity ramp) masks the early frames, so
+the panel materializes already composed and only the last-ranked members visibly land after it —
+and the tile wave carries a small leadIn so no member is mid-fade as the panel edge arrives
+(1f254a158b ("fix(motion): the entrance runs under the slide - the panel materializes
+composed"); a0035941 ("fix(motion): the first tiles wait their turn - lead-in restored, curtain
+dimmed")). The register entry moved from `STAGGER_DECLINED` to `STAGGER_ADOPTERS` with
 the history kept in place — the round trip is the register doing its job, not the refusal being
-overridden. (feat(sidebar): the sections cascade again, gated on the slide the surface now owns.) Take the
+overridden. Take the
 reasoning rather than the removal, because both halves of the complaint were right and only one of
 them was the wave's:
 
--- a/AGENT.md
+++ b/AGENT.md
@@ -4551,7 +4551,15 @@ REFUSED — and a register that only grows is a target.** 9e10b8a9c ("feat(sideb
 sidebar's sections arrive in sequence") gave the right sidebar a wave; the user's verdict was *"I
 don't like the cascading animation effect in the sidebar… This one feels slow. There's a frame drop
 the moment it opens and the moment it closes."* It is off that surface again, and
-`STAGGER_DECLINED` in `tests/test_motion_policy_contract.py` reddens if it comes back. Take the
+`STAGGER_DECLINED` in `tests/test_motion_policy_contract.py` reddens if it comes back.
+**It since came back, the register's way — argued, at the maintainer's request, after both
+refusal facts had been repaired by unrelated work**: the persistent EdgeSlide surface removed the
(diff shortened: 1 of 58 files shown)


XephyLon (commented):
Reviewed at b824910d4 (36 commits, 56 files). Full suite run on the branch in a clean worktree: 1220 QML passed, every harness green, exit 0. CI green. Several claims verified empirically below; several findings are blocking.

## Blocking

**1. The PR cannot be rebase-merged.** The API reports `rebaseable: false`. The branch carries four real merge commits (57e00be54, 6387f8abf, 8d41fb4cb, f011872b4); the last one hand-resolves a conflict in `tests/test_motion_policy_contract.py` (`STAGGER_ADOPTERS`, both sides kept — `git show --remerge-diff f011872b4`). GitHub's rebase-merge replays only the non-merge commits, hits that conflict, and disables the button. Rebase-only is the repo rule: `git rebase gh/main` locally (drops the four merges, resolve `STAGGER_ADOPTERS` once), force-push.

**2. Members hidden at open stay parked at `appear = 0` for the rest of the open.** `StaggerWave.park()` zeroes every member declaring `appear`; `enter()` ranks only members whose `visible` is true and leaves the rest at 0 (`tst_stagger_wave.qml` pins exactly that: "a hidden member spends no slot and stays parked"). Every new adopter with conditionally-visible members inherits it:
- `AiChat.qml`: `suggestions` is `visible` only while `messageInputField.text.length > 0`, `DescriptionBox` only while it has text. Both are hidden at every open (empty composer) → parked → when the user types `/`, the suggestion row and its description become visible at opacity 0. The AI pane's suggestions are dead after this PR.
- `SidebarRightContent.qml` `mediaPlayerLoader`: `visible: active`, `active` gated on a player existing. Open with nothing playing, start playback → the Loader instantiates `Player` (160 px) at opacity 0 — a blank gap in the column until the next open. (The handler-vs-binding order does not race — bindings connect first and `enter()` defers on `pendingEnter` — so this needs playback to begin after the open.)
- `EditModeDrawer.qml` section switch has the same shape but is pre-existing byte-for-byte; noting for completeness only.
`settle()` on close does not fix it — the next open parks again. The wave needs either a visibility-gated park (skip members not visible) or a re-entry on the member's `visibleChanged`.

**3. `Todo.list.indexOf(modelData)` is a no-op on the Gentoo install path.** I probed the identity claim on this machine (`qs -p`, ScriptModel + Instantiator): on the current pin `modelData` IS the list's object — `indexOf` 0/1/2, and a `markDone` diffs as adds=0 removes=1. So the transition claim holds here. But `sdata/dist-gentoo/.../immaterial-impulse-quickshell-git-0.1.0-r6.ebuild` pins `EGIT_COMMIT=7511545` (2026-03-18), which predates quickshell `a611932` "core/scriptmodel: represent members as a QJSValueList" (2026-07-04). At that pin `ScriptModel` stores `QVariant` and compares with `==`, so `modelData` is a fresh object per delegate and `indexOf` returns −1: every done-toggle and delete click is a silent no-op, and the transitions fly the whole list anyway. Fix by resolving the index through something identity-independent (a stable id per task with `objectProp`, or a `findIndex` on content+done), or bump the Gentoo pin — either way say which in the PR.

**4. `tests/test_bar_popup_section_entrance.py` is not wired into `run_tests.sh`.** It is the only `test_*.py`/`lint_*.py` under `tests/` whose name does not appear there (checked all). The body's "nine planted mutations all caught" was a hand run; CI never executes the register. Add the block — and since this is exactly the shape the suite cannot see, a five-line meta-check that every `tests/test_*.py` and `tests/lint_*.py` is named in `run_tests.sh` would make the next one red.

**5. Docs describe a design the branch removed.** AGENT.md (§Design language, the re-adoption paragraph) and the `STAGGER_ADOPTERS` comment for `SidebarRightContent.qml` both say the re-adopted wave "gates on containerProgress through contentsArrived, carries no leadIn". The tree has no `containerProgress` and no `contentsArrived` anywhere under `modules/imi/sidebar*`; `SidebarRightContent` parks and enters on the open's rising edge, ungated, and its own comment says the gated version was measured as ruined (1f254a158). Both texts describe a91c9fc61's design, superseded three commits later. `Docs: updated` is true; updated to the wrong design.

**6. PR body and receipts.** The body covers four streams; the branch has a fifth and the most user-visible one, undescribed: a91c9fc61..b824910d4 — right-sidebar wave re-adopted, EdgeSlide `reveal` curtain on both sidebars (panel ≤15% opacity for the first 40% of travel), convergent tiles, slider fill sweep + icon spin, calendar diagonal ripple, notification status-row converge, left sidebar + AI pane cascade, quick-toggle Behavior removal, new lint. And `Changelog: not user-visible — entries ride the next catch-all release PR` is not true of this PR: everything in it is user-visible motion, and CONTRIBUTING has no "catch-all release PR" step (zero hits) — the receipt names a workflow that does not exist, which is the pasted-receipt shape the rule text warns about. `Changelog: updated`, with `[Unreleased]` entries.

## Should fix

**7. Slider icon rotation is a Behavior stacked on an already-animated value.** `QuickSliders.qml`: `rotation: quickSlider.value * 360` plus `Behavior on rotation { NumberAnimation 350ms OutBack }`. `value` already glides through the slider's `SmoothedAnimation`, so rotation is smooth without the Behavior; with it, every frame of the 650 ms sweep (and of any drag) retargets the Behavior — b710ef731's per-frame-restart shape, the same one this PR's new lint fails on for opacity. The review agent measured it: the icon sits at 0° for the whole sweep, then whips the full turn in 350 ms with an 18° overshoot — the opposite of "spins with the fill". Drop the Behavior.

**8. Calendar ripple replays on every tab switch.** `BottomWidgetGroup.onLoaded` installs `entranceTrigger` as a late `Qt.binding`; the re-created `CalendarWidget` starts at −1 and takes the current count → `onEntranceTriggerChanged` fires on every Todo→Calendar switch, replaying the ~530 ms diagonal ripple over the 200 ms tab fade. Seed the initial value before binding, or ignore the first edge.

**9. `keepRightSidebarLoaded = false` loses the whole right-sidebar entrance.** With that (user-settable, Settings > Sidebars) toggle off, `sidebarContentLoader.active` follows `slide.shown` → `GlobalStates.sidebarRightOpen` synchronously, so the content — and its `Connections { target: GlobalStates; function onSidebarRightOpenChanged() }` blocks in `SidebarRightContent.qml` and `AndroidQuickPanel.qml`, the only drivers of `park()/enter()` and `entranceTrigger++` — is created *inside* the `sidebarRightOpenChanged` emission and never receives that edge (`QMetaObject::activate` skips connections made after the emission's `highestConnectionId` snapshot; reproduced under qmltestrunner on Qt 6.11 and 5.15, with a pre-loaded control that does fire). Torn down on every close, so it misses every open: no section wave, no tile wave, no sweep, no ripple, no converge — nothing visibly broken, the feature is just absent. The left sidebar is unaffected (content created once at boot). Drive the entrance from something the content can observe after creation (e.g. `Component.onCompleted` when already open, or the slide's own `progress`).

**10. Slider sweep windows.** `QuickSliders.qml`: `valueVelocity = Math.max(0.05, shownValue / 0.65)` is a timed override restored ~730 ms later, and the park is a binding flip that rides the same `Behavior on value`. (a) A zero reading at open (volume 0, no sink, mic 0, brightness at the gamma floor — not mute) installs 0.05 u/s, so a volume key pressed inside the window moves the fill by nothing, then snaps when the binding is restored. (b) Close and reopen inside the window: `entranceParked` drives `value → 0` through the still-installed sweep velocity — a visible ~220 ms dip, then the second sweep reverses it. Snap the park (bypass the Behavior) and re-derive the velocity on `shownValue` changes, or drop the velocity trick for a dedicated animation.

**11. A tile added mid-session pops in with no fade.** `AndroidQuickToggleButton.qml`: the removed `opacity: 0` + `onCompleted` self-fade was the only entrance a delegate created *after* the open wave had; `StaggerEntrance` binds a late arrival straight to `appear = 1`. Edit mode: click an unused tile → new delegate in `usedGrid` at full opacity in one frame; the reverse path lands in `unusedGrid`, which has no wave at all. The diff's own comment concedes it ("simply appears in place"). A late-dressed member should start at 0 and run one fade.

**12. `wavePending` leaks across a pre-gate takeover** (`BarPopupOverlay.qml`): armed only from idle, cleared only in `onContentsInChanged`/`finishExit`, not `beginExit`. Hover A, slide to B within ~90 ms → the gate fires against B's never-parked tree and `enter()` snaps every B section to 0 then cascades — contradicting the "cross-fade arrives at full strength" comment. Mostly masked by `contentEnter` holding B at opacity 0 for ~100 ms; a visible blink only in a same-frame race. Clear the flag in `takeOver`'s cross-fade branch.

## Non-blocking

- **Blur under the curtain.** `SidebarRight.qml` gates `WindowBlurRegion.regionItem` on `slide.shown` (true from frame one) while the body sits at ≤0.15 opacity for 40% of the travel — the compositor frosts the full rect under a ghost panel, ~140 ms per open and close. The tree names that shape in the overview's blur contract. I did not capture it (no recording without a go-ahead); worth one 60 fps look if you have not measured it.
- **Raw constants beside the tokens this PR added.** `NotificationList` converges from `±40` px while `Appearance.animation.convergeReachX` is the token for that reach; `CalendarDayButton` `(row+col)*28`/220 ms, `AiChat` 120/380 ms + overshoot 1.2, sliders `80+50i`/650/350 ms + overshoot 1.5 — all on `Appearance.animation.scale()` but none on a tier, so the tier lints cannot see them. Promote (as `convergeReach` was) or record the exception.
- **Two entrance derivations.** `bar_popup_unroll.js` `entranceScale`/`entranceOffset` and `motion_policy.js` `entranceScaleFrom` carry the same floor and the same rise separately, with no test tying them; `StaggerEntrance` cannot replace the bar copy yet because it never re-dresses on a `target` change (the overlay's target swaps per popup).
- **Seven spellings of arm/run.** Four near-identical `Connections { target: GlobalStates … park(); enter() }` blocks plus three different container gates (overlay, drawer, desktop menu), settle-on-interrupt at only three sites; `enter()` already writes `appear = 0`, so `park(); enter();` is itself a duplicated idiom. `StaggerWave` could own "enter on this flag's rising edge" once.
- **`EdgeSlide.reveal` is unpinned.** The one entrance curve in this PR living in a rendering widget with literal 0.4/0.15 instead of `motion_policy.js`, and nothing in `tests/` names the numbers — unlike `entranceScaleFrom`/`convergeFrom`/`convergeSettle`.
- **M3 §2 / §Stagger.** `docs/M3_GUIDELINES.md` forbids raw durations, `Easing.OutCubic`, and `index * <ms>` ladders outside `staggerDelay()`; all eleven `Appearance.animation.scale(<literal>)` sites outside `Appearance.qml` are new in this PR (`CalendarDayButton`, `NotificationList`, `AiChat`, `QuickSliders`). `NotificationList`'s 350 ms is `expressiveFastSpatial`'s duration without its curve — the "different tier's number" smell AGENT.md names; `convergeSettle` is the house spelling for that converge.
- `ai_catalog.js` `resolve()` evaluates `legacyId.length`; `buildModel` copies a `null` legacyId through → `TypeError` once `extraModels` feeds it (stage 2). `typeof === "string"` guard.
- Comment in `lint_wave_member_opacity_behavior.py` cites `8a0120c2` bare — dangles after the rebase (the citation lint only covers AGENT.md/CONTRIBUTING, so it goes quietly stale).

## Verified OK
- ScriptModel identity + delegate `modelData` on the current Arch pin (probe above).
- The two SHA citations in AGENT.md carry subjects, and `lint_doc_citations.py` resolves by subject after a rebase.
- The dresser replaces no call-site binding: `DescriptionBox`, `VerticalTabBar`, `NotificationStatusButton`, `GroupButton` roots bind neither `opacity` nor `scale`.
- Three of the four merges are clean remerges; only f011872b4 carries a resolution.

**Verdict: not mergeable as-is** — 1 (rebase), 2, 3, 4, 5, 6 block; 7–12 should land in the same round. Re-review on request.

XephyLon (commented):
Re-reviewed at 22a8ad773 (rebased, 0 merges, `rebaseable: true`; 12 fix commits on top of the reviewed set). Full suite in a clean worktree: 1220 QML passed, 206 harnesses, exit 0 — including the newly wired `test_bar_popup_section_entrance.py` and `lint_suite_registration.py`.

## Blocking, from the first round

1. **Rebase** — done.
2. **Hidden members parked** — done: `enter()` settles unranked members to 1; `tst_stagger_wave` pins the new contract. A member shown mid-open arrives at rest rather than fading, which is the pre-wave behaviour and the right trade.
3. **Gentoo `indexOf`** — done: `resolveIndex()` tries identity, falls back to `findIndex(content+done)`, both handlers bail on −1; the test now requires both halves.
4. **Unwired register** — done: wired, plus `lint_suite_registration.py`. Nit: the runner comment says "checked before anything else" and "the meta-check at the end of this file"; it sits at line ~962 of ~1000, neither first nor last.
5. **Docs described the removed design** — done: AGENT.md and the `STAGGER_ADOPTERS` comment now describe the ungated, curtain-masked design; citations carry subjects, so they resolve after the merge rewrites SHAs.
6. **PR body** — **not done.** Still four streams (the sidebar re-adoption / curtain / converge / sweep / ripple stream — a91c9fc61..22a8ad773's biggest and most user-visible — is undescribed), and the receipt still reads `Changelog: not user-visible — entries ride the next catch-all release PR, per the release workflow`. CONTRIBUTING §Keep CHANGELOG.md fed says the opposite (the PR that earns the entry writes it; a release reconstructing from the log is the named failure). Whether that section or the practice changes is the maintainer's call; until one of them does, the receipt contradicts the written rule.

## Should-fix, from the first round

7. rotation `Behavior` — removed. 8. calendar replay — armed one turn after creation; ordering against the parent's `Qt.callLater` is right (children complete first). 9. `keepRightSidebarLoaded = false` — both drivers run the entrance from `Component.onCompleted` when born open. 10. slider windows — park lands with the glide off, no velocity override near zero, a reading change ends the sweep. 11. late tile — parks itself and runs one fade, stands down when a wave is active or armed; a born-open tree still gets the wave because `requestSync`'s `callLater` is queued before the wave's. 12. `wavePending` — cleared in the takeover branch.

All read as correct; none re-measured on screen.

## Non-blocking

- `ai_catalog` null `legacyId` and the `8a0120c2` citation — done.
- New: `test_motion_policy_contract.py` cites `1f254a158b` bare (twice), a pre-rebase SHA with no subject — not covered by the citation lint, dead once the old head is garbage-collected. Add the subject or drop the SHA.
- The rest of the first round's non-blocking list (blur under the curtain, raw constants beside the tokens, two entrance derivations, seven arm/run spellings, `EdgeSlide.reveal` unpinned, M3 §2) is unchanged; fine to leave, worth a line in the body so it is not rediscovered.

**Verdict: mergeable once the body is fixed (6)** — describe the fifth stream, and either write the `[Unreleased]` entries or settle the CONTRIBUTING question first. Code is done.

PR #351, opened 2026-09-04
feat(wallpaperSelector): the Wallpaper Engine sidebar — settings, properties, compatibility, crop picker, Quality; plus resource-polling perf

XephyLon:
The wallpaper selector gains a Wallpaper Engine sidebar, and the resource-polling
perf work it grew on. Scope is the 13 commits on this branch over `main` (rc-12);
the widget-options / shadow-switch / rc-11–12 work an earlier draft mentioned is
already on `main` and not in this diff.

## Wallpaper Engine selector
- **A sidebar** — a places rail over local files, the engine's config over WE
  wallpapers (modelled on jagrat7/linux-wallpaper-engine).
- **The engine's full flag set + per-wallpaper properties + a compatibility
  scan** — frame rate, scaling, audio, volume, the audio-reactive recorder,
  mouse/parallax/particles, and each wallpaper's project.json properties
  (bool/slider/combo/color/textinput), global or per wallpaper. A spawned
  scanner marks broken wallpapers.
- **A Fill crop picker** on the active wallpaper's thumbnail — pick which slice
  of an over-sized scene is shown; the renderer pans it live (focus read every
  frame, no reload). The thumbnail is the real scene grabbed from the renderer
  (a 32:9 scene reads as 32:9, upright), not the portrait preview.
- **A render-scale Quality dial** — Native / High / Balanced / Low, gated on a
  renderer that reads it, and honest that fps (not scale) is the lever for a
  heavy scene.

Every extended surface property binds in-guarded, so a shell on an older
qs-wallpaperengine binary hides the controls it cannot honour instead of
breaking. Needs the matching module surface, **XephyLon/qs-wallpaperengine#21**
(merged), for users to see the new controls once `WE_REF` moves to a release
carrying it.

## Performance (the branch's origin)
- **Resource polling reads through files** and stops waking a sleeping dGPU:
  nvidia-smi runs only when the card is not runtime-suspended, and — after this
  review — backs its cadence off once the card reads idle, so an *active* card
  is also allowed to autosuspend instead of being held awake by the poll itself.
- **Wallpaper decode bounded to what a screen can draw** (now × devicePixelRatio,
  so fractional-scale outputs stay sharp); the session menu's surface outlives
  its gesture; the avatar resolves once instead of retrying a guessed path.

## Review round (maintainer review addressed)
HiDPI decode bound, the nvidia-smi autosuspend backoff, the compat scanner
blaming the queue head only on a real crash (and a window rule so its scanner
window never tiles/steals focus), color-property serialisation (always a dot, so
WE does not read white as near-black) and its 0..255 parse, combo numeric option
keys, the volume slider re-binding after a drag, `ignoreUnknownSignals` on the
0.3-only signals, the scene grab labelled and gated at request time, suspended
GPU readings zeroed, nvidia detection requiring a real 0x10de device, `sanitize`
typing every flag, a corrupt override store quarantined before overwrite, and
the AGENT.md citation/contradiction fixes.

Docs: updated AGENT.md §WallpaperEngineFeatures.qml (the Quality dial replaces the "not shipped" note) and citation fixes across the WE/resources entries.
Changelog: updated

### DIFF
--- a/AGENT.md
+++ b/AGENT.md
@@ -908,9 +908,17 @@ services/                  Singletons wrapping external state/processes - one pe
   WallpaperEngineFeatures.qml  Which properties THIS binary's embedded renderer has, probed once
                               by building a bare WallpaperEngineSurface (no project, so no thread
                               or GL) and asking `"x" in surface`. The extended controls (the
-                              qs-wallpaperengine 0.3 flag set, per-project properties) gate on
-                              it; WallpaperEngineLayer binds the same properties dynamically for
-                              the same reason. See "WE_REF pins the renderer" below
+                              qs-wallpaperengine 0.3 flag set, per-project properties, the live
+                              crop focus) gate on it; WallpaperEngineLayer binds the same
+                              properties dynamically for the same reason. See "WE_REF pins the
+                              renderer" below. The Fill crop picker
+                              (modules/imi/wallpaperSelector/crop_picker.js +
+                              WallpaperSelectorSidebar's active card) reads
+                              GlobalStates.weContentAspect - the LIVE scene's real authored
+                              aspect, published by Background from the surface's contentWidth/
+                              Height, because a 32:9 wallpaper's portrait preview would pick the
+                              wrong overflow axis - and drives focusX/focusY, which pan the
+                              wallpaper live (scene only; a video is centre-cropped by WE)
   WallpaperEngineCompat.qml    Per-wallpaper compatibility verdicts, the reference app's bulk
                               scanner: the scan runs in a SPAWNED `qs -p` scanner process
                               (scripts/wallpapers/we_compat_scan.qml) loading each project into a
--- a/AGENT.md
+++ b/AGENT.md
@@ -871,6 +871,21 @@ services/                  Singletons wrapping external state/processes - one pe
                               takes to appear, the original persisted in Persistent so a
                               restart mid-swap still undoes it - see the entry under
                               "External binaries the shell drives"
+  WallpaperEngineOverrides.qml Per-project Wallpaper Engine settings (fps, scaling, audio), keyed
+                              by runtime project ids so it is a raw FileView over
+                              wallpaper-engine-overrides.json on the PluginState pattern - never
+                              a JsonAdapter. services/we_overrides.js is the resolution (per-key
+                              fallback to the globals, clear-on-null, sanitation); `active` is
+                              THE one live derivation, read by WallpaperEngineLayer's
+                              fps/scaleMode, Background's audio routing and the selector
+                              sidebar's controls - tests/test_we_overrides_wiring.py refuses a
+                              raw-config read beside it. The sidebar
+                              (modules/imi/wallpaperSelector/WallpaperSelectorSidebar.qml) is the
+                              editor: a places rail over local files (selector_places.js), the
+                              engine's configuration over Wallpaper Engine, and only the knobs
+                              WallpaperEngineSurface actually answers - a control for a flag the
+                              renderer does not read would be a fake action
+                              ("feat(wallpaperSelector): a sidebar - places over files, engine config over WE")
   SchemePreview.qml            Per-scheme swatches for the scheme pickers: one venv run of
                               scripts/colors/scheme_preview.py quantizes the wallpaper once and
                               builds every Material variant from it. Cached against the wallpaper
--- a/AGENT.md
+++ b/AGENT.md
@@ -6559,8 +6559,15 @@ map-time keyboard grant. A third obligation lives in `services/GlobalFocusGrab.q
 the grab's whitelist must end with the dismissables — Hyprland hands the grab's keyboard focus to
 a surface picked from the END of the list, and with the bar/OSK after the panel, an opening panel
 never activates and every keypress is silently dropped (the persistent-sidebar contract pins the
-order; fix(sidebars): route the focus grab's keyboard to the opening panel). `SessionScreen` still
-maps per open and pays this same 61ms.
+order; fix(sidebars): route the focus grab's keyboard to the opening panel). `SessionScreen` is
+that family too now — per-screen windows, the latch, Exclusive keyboard and the whole-surface
+mask both gated on `isTarget`, an exit that owns `reallyOpen` — with one addition of its own:
+`rules.lua`'s `quickshell:session` rule was `ignore_alpha = 0`, under which a permanently-mapped
+surface's fully transparent idle pixels would ask the compositor to blur the entire screen, so
+the threshold moved to 0.4 (the scrim sits at ~0.88+, so it frosts exactly as before). Its layer
+follows the overview's #339 rule — Overlay only while open over a true fullscreen window, Top
+otherwise — because a permanently-mapped Overlay surface holds the fullscreen fast path shut.
+("perf(sessionScreen): the session menu's surface outlives the gesture").
 
 A fourth cost, found by the first multi-monitor user to update (#297): **a persistent surface has
 to say which screen it lives on.** A `PanelWindow` with no `screen:` asks the compositor to choose
--- a/AGENT.md
+++ b/AGENT.md
@@ -918,7 +918,14 @@ services/                  Singletons wrapping external state/processes - one pe
                               aspect, published by Background from the surface's contentWidth/
                               Height, because a 32:9 wallpaper's portrait preview would pick the
                               wrong overflow axis - and drives focusX/focusY, which pan the
-                              wallpaper live (scene only; a video is centre-cropped by WE)
+                              wallpaper live (scene only; a video is centre-cropped by WE). The
+                              picker draws the actual scene, grabbed by the renderer
+                              (requestSceneGrab -> GlobalStates.weSceneGrabPath, a PNG at the
+                              real 32:9 aspect), NOT the preview image - the preview does not
+                              depict an ultrawide scene. Note: WE renders a scene at its authored
+                              resolution regardless of window, so a render-scale lever does
+                              nothing for scenes (measured) and was not shipped; fps is the
+                              scene's only in-shell cost lever
   WallpaperEngineCompat.qml    Per-wallpaper compatibility verdicts, the reference app's bulk
                               scanner: the scan runs in a SPAWNED `qs -p` scanner process
                               (scripts/wallpapers/we_compat_scan.qml) loading each project into a
--- a/AGENT.md
+++ b/AGENT.md
@@ -367,7 +367,7 @@ at startup). None of the documented env vars stops the probe
 (`QT_FFMPEG_DECODING_HW_DEVICE_TYPES` filters codec selection, not this enumeration). Before
 adding a QtMultimedia object anywhere, know that its first construction costs a near-second stall
 and permanently attaches the process to the dGPU.
-7ef4e762 ("perf(resources): poll through files, and stop waking a sleeping dGPU").
+417e9819 ("docs(agent): the first QtMultimedia object probes every hw context, cuda included").
 
 **A sound event is one `pw-play` on a path the shell resolved, and neither half of that sentence
 was true before.** `Audio.playSystemSound()` built
@@ -904,7 +904,7 @@ services/                  Singletons wrapping external state/processes - one pe
                               Wallpaper Engine, gated on WallpaperEngineFeatures so an older
                               renderer binary shows only what it answers - a control for a flag
                               the renderer does not read would be a fake action
-                              ("feat(wallpaperSelector): a sidebar - places over files, engine config over WE")
+                              c0c05d3f ("feat(wallpaperSelector): a sidebar - places over files, engine config over WE")
   WallpaperEngineFeatures.qml  Which properties THIS binary's embedded renderer has, probed once
                               by building a bare WallpaperEngineSurface (no project, so no thread
                               or GL) and asking `"x" in surface`. The extended controls (the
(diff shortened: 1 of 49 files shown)


XephyLon:
Full read of the 12 commits, plus the failing check. Grouped by what blocks merge, what is a real defect, and what is a nit.

### Blocking

1. **CI `receipt` fails: no `Changelog:` line, and the diff never touches `CHANGELOG.md`.** The whole WE sidebar, the crop picker, the Quality dial, the dGPU polling change and the session-surface change are all user-visible and none are in the changelog. Needs `Changelog: updated` and real entries under Unreleased.
2. **The body is stale.** It says the per-surface widget options, the shadow switch and rc-11/rc-12 "ride on this branch" and that the `Docs:` receipt covers "CHANGELOG.md (rc-11/rc-12 entries) on this branch". The merge base is `890b5d89c` (rc-12) and none of that is in the diff. Rewrite to what the 46 files actually change.
3. **`AGENT.md` contradicts the code it documents.** The `WallpaperEngineFeatures.qml` entry says a render-scale lever "does nothing for scenes (measured) and was not shipped; fps is the scene's only in-shell cost lever". bc571a8d6 ships the Quality dial. Update the entry (the dial's own honest caveat, scene vs video, is the right text).
4. **`AGENT.md` cites a commit that does not exist.** `("feat(wallpaperSelector): a compatibility scan that marks the wallpapers the renderer cannot start")` matches nothing on the branch; the compat scan landed in dd97c0a16 "the engine's full settings, per-wallpaper properties, and a compatibility scan". Four other new citations are the bare-subject form with no SHA (`("perf(background): …")`, `("perf(sessionScreen): …")`, `("feat(wallpaperSelector): a sidebar …")`, and the one above), which is exactly the form `lint_doc_citations.py` does not check, so the lint passed. Use `<sha> ("<subject>")` for all of them. Also `AGENT.md:364`: the QtMultimedia hw-context paragraph cites 7ef4e762 (resource polling); the commit behind it is 417e98199.

### Defects

5. **HiDPI regression in the wallpaper decode bound** (`Background.qml` `wallpaperDecodeSize`, 61aec9097). `ShellScreen.width/height` are logical pixels; `Image.sourceSize` is decoded pixels. On a scale-2 (or 1.25/1.5) output the wallpaper, the frost and the depth cutout now decode at a fraction of the panel's pixels and get upscaled by the scene graph: visibly softer for every fractional-scale user. The commit's measurements were all at scale 1, and nothing in the shell reads `devicePixelRatio`, so nothing caught it. Fix: multiply both axes by `bgRoot.modelData.devicePixelRatio`; the sharers inherit it through the same property.
6. **The `nvidia-smi` gate preserves a suspended dGPU but never lets one suspend** (`ResourceUsage.qml` fast tick, 7ef4e7629). Once `runtime_status` reads `active` for any reason (a game, the CUDA probe, boot), the 3 s poll resumes, and by the new AGENT.md paragraph's own reasoning each wake outlasts the driver's idle window, so the card stays awake for the rest of the session. Net effect on a hybrid laptop: suspend survives only if it was already there at shell start. Fix: while `active`, back the `nvidia-smi` cadence off to at least `power/autosuspend_delay_ms` (readable next to `runtime_status`), or poll every k-th tick after N zero-utilisation readings. `nvidiaShouldPoll` and the contract test stay as they are.
7. **Cross-repo: vertical crop focus.** The shell sends `focus.y = 0` for the top of the picture (`crop_picker.js` `viewport`/`focusFromPointer`, and the tests pin it). That is the right convention, but the module in XephyLon/qs-wallpaperengine#21 inverts `focusY` in the fill blit (see the comment there), so on the current #21 head every off-centre vertical drag pans the wrong way. Nothing to change here; this PR should not merge ahead of the #21 fix.

8. **Color properties serialise to a form the engine reads as 0..255 ints** (`WallpaperPropertyControl.qml`, the color commit). `Math.round(c * 1000) / 1000` joined by spaces has no `.` whenever every channel rounds to an integer: white is `"1 1 1"`, pure red `"1 0 0"`. The vendored engine's `ColorBuilder::parse` treats a dot-less triplet as 0..255 integers, so white becomes (1/255, 1/255, 1/255), near black. Fix: `channels.map(c => c.toFixed(3)).join(" ")` (always carries a dot) and pin it in a test. Same family, `channel()` in the swatch: a project default in int form (`"255 128 0"`, which the engine accepts) reads as (1, 1, 0) because it clamps to 0..1; divide by 255 when no channel string contains `.`.
9. **The compat scanner blames the queue head for any silent exit, including "could not start"** (`WallpaperEngineCompat.qml`, the process-exit handler). A scanner that never gets going (stock binary without `Quickshell.WallpaperEngine`, exec failure, oversized env) exits with no marker and no progress, so the head is recorded broken, the respawn marks the next head, and so on until `respawnsLeft` runs out: a persisted false "broken" for the whole front of the queue. Only blame by position on `exitStatus === QProcess.CrashExit` (a renderer crash is a signal death; a startup failure is `NormalExit` with a non-zero code); on a normal non-zero exit with no output, stop the scan and record nothing.
10. **The scanner's `FloatingWindow` is a visible, focus-stealing tile** (`we_compat_scan.qml`, title "Wallpaper compatibility scan"). No rule for it in `rules.lua` (the diff only touches `quickshell:session`), so Hyprland tiles it into the active workspace and focuses it on every spawn and respawn, once per broken wallpaper in the worst case. Needs a window rule on that title: `float`, `no_initial_focus`, and `workspace special:silent` or an offscreen position.

### Nits

- `WallpaperSelectorSidebar.qml` volume slider: `value: root.effective.volume` is a plain `Slider` binding, broken by the first drag. After one drag, flipping Custom settings or switching projects leaves the handle where it was. Same latent pattern as `ConfigSlider`; a `Binding` on `value` or re-assigning on `effective` change fixes it.
- `Background.qml`: the `Connections` for `onContentSizeChanged` and `onSceneGrabReady` need `ignoreUnknownSignals: true`. Qt checks handler-vs-signal at connect time regardless of `enabled`, so an older renderer binary logs "no signal of the target matches" per screen.
- `Background.qml` `onSceneGrabReady` labels the grab with `activeProject` at ready time, not request time. Capture the id when the grab is requested so a switch during the 400 ms + render window cannot mislabel the previous scene.
- The grab re-runs on every `rendered` edge, which includes every reload (a volume commit, a Quality change, a property edit). Each is a full-scene `glReadPixels` + PNG encode on the render thread. Fine for now; gating it on `canCrop` (Fill, overflowing, renderer supports focus) would skip most of them.
- `ResourceUsage.qml`: on `suspended`, only `gpuUsage` is zeroed; `gpuTemp`/`vramUsed` keep the last awake values, so the card shows a stale temperature for a card that is off.
- `ResourceUsage.qml`: vendor is `nvidia` purely on `command -v nvidia-smi`; the probe already walks PCI for 0x10de, so require the hit before choosing `nvidia`, or a CUDA-userland AMD box polls a failing `nvidia-smi` forever.
- `we_overrides.js` `sanitize` type-checks only `renderScale`; a hand-edited `"fps": "60"` or `"silent": "no"` survives and reaches `WallpaperEngineSurface.fps` as a QML type error on every re-evaluation. Coerce or drop the other flags the way `renderScale` is handled.
- `WallpaperEngineOverrides.qml`: an unreadable store logs and continues with `{}`; the next `setOverride` overwrites the corrupt file with one record, losing whatever the user had. Rename the bad file aside before the first write.
- `wallpaper_engine.py` combo options: values are emitted `str(value)`; the engine keys numeric options by `std::to_string(int)`, so a float option (`1.5`) never matches and the override is rejected. Mirror the int cast for numeric values.
- Follow-up outside this PR: nothing here reaches a user until qs-wallpaperengine tags a release with #21 and `WE_REF` in `sdata/subcmd-install/4.wallpaperengine.sh` (currently `v0.2.6`) moves to it.

### Checked and sound

Every extended surface read goes through an `"x" in root` guard (`WallpaperEngineLayer.qml`) and the sidebar gates on `WallpaperEngineFeatures`, so an older binary degrades to the old controls. `crop_picker.js` viewport/focus maths round-trip and the tests cover both axes, clamping, and no-overflow. The one-writer rule for `weContentAspect` (greeter-still owner) and the reset on `weLoader.active` false hold. `rules.lua` `ignore_alpha = 0.4` sits under the ~0.88 scrim. SessionScreen's `isTarget` gating, layer toggle and `reallyOpen` ownership mirror the overview's shape. AMD/Intel sysfs parsing keeps units and the zero-total guard; FileView's empty first read is handled on every branch. Avatar precedence is a strict improvement and every consumer imports `qs.services`. `WallpaperEngineGrid` type normalisation and the Hide broken filter both read through `statusFor`. The rest of the `--set-property` contract matches the engine (`name=value` split at the first `=`, bool `"1"/"0"`, slider via `stof`, a bad value throws inside project setup and surfaces as `failed`, not a crash). Store semantics hold: null never persisted, centre focus stored as absence, empty records dropped, `clearEngineOverrides` leaves `properties`/`focus` alone, keys are Workshop ids so paths never enter the map. Scanner hygiene: one project at a time, bounded respawns, per-project deadline plus a silence watchdog, dies with the singleton on reload. Config defaults reproduce the renderer's old fixed behaviour. The three new Python contracts are registered in `run_tests.sh`, stdlib only, each with a can-fail self-test; the `tst_*.qml` files are picked up by the runner.

XephyLon:
All of it addressed (commits `ac57b901`, `d4219bf0`, `1057c0f7`, `c9a8ffbb`, on top of the changelog/body fixes).

### Blocking
1. **Changelog receipt** — `[Unreleased]` now carries the WE sidebar, crop picker, Quality dial and the dGPU-wake fix; body has `Changelog: updated`.
2. **Stale body** — rewritten to the 13 commits actually in the diff (merge base rc-12); the widget-options/rc-11–12 claims are gone.
3. **AGENT.md contradiction** — the WallpaperEngineFeatures.qml entry now describes the shipped Quality dial with its scene-vs-video caveat, not "not shipped".
4. **AGENT.md citations** — the four bare-subject citations carry their SHAs; the compat-scan citation is `dd97c0a1` with its real subject; the QtMultimedia paragraph now cites `417e9819`, not `7ef4e762`. Lint green.

### Defects
5. **HiDPI decode** — `wallpaperDecodeSize` × `devicePixelRatio`; the frost and cutout inherit it through the same property.
6. **nvidia-smi backoff** — after N idle readings the poll drops to once every k ticks so an active-but-idle card can autosuspend; load snaps back to full cadence. `nvidiaDueForPoll` is unit-tested; the gate contract still holds.
7. **focusY (cross-repo)** — fixed and merged in #21; this branch's `crop_picker.js` convention was already right. Not merging ahead of it.
8. **Color serialisation** — always a dot (`toFixed(3)`) so white isn't read as near-black; dot-less defaults parsed as 0..255. Extracted to `we_color.js` + `tst_we_color.qml`.
9. **Compat blame** — position-blame only on `CrashExit`; a normal exit with no marker stops the scan and records nothing.
10. **Scanner window** — Hyprland rule: float, `no_initial_focus`, parked offscreen (stays mapped so `rendered` still works).

### Nits
- Volume slider re-binds via a `Binding` element (survives the first drag).
- `ignoreUnknownSignals` on the contentSizeChanged/sceneGrabReady Connections.
- Grab labelled and gated at request time; also kicked from `onContentSizeChanged`, coalesced by the 400 ms Timer, so neither signal order misses it.
- Grab gated on `canCrop`, so a non-croppable reload doesn't pay for a `glReadPixels`.
- `gpuTemp`/`vramUsed` zeroed on suspend.
- nvidia chosen only with a real 0x10de device on the bus.
- `sanitize` types every flag now, not just `renderScale`.
- Corrupt override store quarantined (`mv -n`) before the next write.
- Combo numeric option values int-keyed; slider values stay float.

Full suite green (the one failure on the run was `WeatherPopupHeroRuntimeTest`, a bar-growth animation harness unrelated to this diff — passes isolated).

XephyLon:
Round 2 (8efa2cc1c..c9a8ffbb6) read in full. Every blocking item and every defect from the first pass is addressed, and the fixes match what the review asked for:

- Receipts present, CHANGELOG entries added, body rewritten to the real scope. All three checks green.
- AGENT.md: all five citations now `<sha> ("subject")` and resolve; the Quality-dial entry matches bc571a8d6; the QtMultimedia paragraph cites 417e98199. `lint_doc_citations.py` passes on the head.
- `wallpaperDecodeSize` multiplies by `devicePixelRatio`; the sharers inherit through the property.
- nvidia: vendor requires a 0x10de display device; idle backoff after 3 zero readings polls once per ~12 s; suspended zeroes temp and VRAM too. `tst_resource_usage` pins one poll per window.
- Compat scanner: position blame only on `CrashExit`; a normal exit with no marker stops and records nothing; window rule floats, denies focus and parks it offscreen.
- Color: `we_color.js` always writes a dot (`toFixed(3)`), reads dot-less as 0..255; round-trip test covers white. Combo option values integer-keyed, slider floats kept, tested.
- Volume slider re-asserted through a `Binding`; `ignoreUnknownSignals` on both 0.3-only `Connections`; grab labelled at request time and gated on `weCanCrop`; `sanitize` types every flag; corrupt store quarantined once per session before any write.

Nothing left that should hold the merge. Three optional follow-ups, none of them blocking:

- The backoff gap is a fixed ~12 s. `power/autosuspend_delay_ms` sits next to `runtime_status` and is readable for free; a user who raised the delay past 12 s never suspends under this poll. Using the file makes the gap exact instead of generous.
- `nvidiaZeroStreak` is not reset when the gate reads `suspended`, so the first busy sample after a wake (a game start) can lag up to one backoff window. Resetting the streak in the suspended branch makes a wake a fresh episode.
- A compat scanner that fails to start now stops silently; the button just returns to "Check compatibility". A one-line error on the service (shown where the active verdict row is) would tell the user why nothing was tested.

Outside this PR: qs-wallpaperengine#21 is merged but untagged (latest tag v0.2.6), and `WE_REF` still pins v0.2.6, so the new controls stay hidden for every user until a release is cut there and the pin moves.

PR #352, opened 2026-09-04
fix(media): lyrics fetch caching + a shared MediaArtSource, and the media/lyrics review round

XephyLon:
A review of the media sidebar and the lyrics feature, then the fixes it surfaced.

## Lyrics — fetching (the headline)
- **On-disk cache.** `lyrics.py` re-walked up to five providers (each a 12s urlopen, plus Glassy's ~9s CDP poll) on every fetch. It now caches the answer under `$XDG_CACHE_HOME/immaterial-impulse/lyrics` keyed by normalized (title, artist, duration) — positive 30 days, `not_found` 6 hours, bounded to 512 entries. A replay or a sidebar reopen is now instant.
- **Debounce + dedup.** Metadata lands field by field and a track change fires several signals; each launched, killed, and relaunched the fetcher. The restart is debounced to one fetch, deduped against the last key.
- **Watchdog.** A hung fetch ends after 20s instead of spinning the spinner forever.

## Lyrics — bugs
- Seeking called `lyricsComp.restartLyrics()`, which doesn't exist on `Lyrics` — a `TypeError` every drag; and re-fetching on a seek is wrong (the sync clock re-anchors itself). Removed.
- A deliberate fetch-kill's late `onExited` flipped the fresh load to `not_found` — the "spinner → no lyrics" flicker on every track change. Guarded.
- A `Lyrics` view now claims the service's single `overridePlayer` only while visible and releases it on destruction, so a hidden/gone view doesn't strand the service on a stale player.
- An enhanced-LRC line repeated at several timestamps mis-timed its karaoke (one word-stamp set, absolute to the first position, attached to all). Repeated lines drop their words.

## Media sidebar / controls
- **One shared `MediaArtSource`** replaces the cover-art pipeline copy-pasted across four surfaces — which had drifted: the sidebar lacked `curl -4` (IPv6 stall) and the `file://` fast-path. 
- **DockMedia tint froze grey** after an art-less track (imperative write to a bound property — the regression its siblings warn against). Fixed by the shared source.
- **Progress bars divided by zero/absent length** (NaN with no player, Infinity on a live stream) into the slider value; guarded everywhere.
- **Perf:** the left sidebar's position poll ran all session while hidden (gated on `sidebarLeftOpen`); DockMedia's blur decoded art full-res (bounded to the card); `WaveVisualizer` reallocated its buffer every paint (reused).
- Dead code removed; two unguarded player derefs made optional.

New/updated tests: the lyrics cache (roundtrip, TTL, main-served-from-cache), the multi-stamp fix, all against stubbed IO. Full suite green; shell loads clean.

Note on scope: two review findings were deliberately NOT taken — `desktopWidgetLyricsActive`/`slots` are pinned by `test_lyrics_widget_contract.py` as intentional API, and routing transport through `MprisController` would break the sidebar's per-player control.

Docs: updated AGENT.md §"two names for one thing" (MediaArtSource is the one cover-art pipeline).
Changelog: updated

### DIFF
--- a/AGENT.md
+++ b/AGENT.md
@@ -1925,6 +1925,17 @@ ce41c4f9c ("feat(cava): give CavaService the producer it always implied"),
 bcf5f9ca1 ("refactor(cava): move every band consumer onto the one service"),
 004a17745 ("test(cava): pin the producer, the gate and the band contract").
 
+The same rule holds for cover art: **the download+quantize pipeline lives once, in
+`modules/imi/mediaControls/MediaArtSource.qml`** - feed it `artUrl` (via `media_art.js`'s
+resolve), read `displayedArtFilePath` and `colors`; each surface keeps only its own
+dominant-colour mix. It was copy-pasted across the bar, dock, left sidebar and popup players, and
+the copies had drifted exactly as CavaService's consumers did: the sidebar's lacked the `curl -4`
+IPv4 pin (its downloads could hang on an IPv6 stall) and the `file://` fast-path (it re-curled
+local art every track). A fifth media surface must use this component, not paste a fifth copy.
+It cannot live in `modules/common/widgets` - it spawns curl, and `lint_dumb_widgets.py` fails the
+suite on a Process there.
+1550313c ("refactor(media): one shared MediaArtSource; fix tint, progress, and waste").
+
 **A signal nothing connects to is that same hole from the other side, and only a surface can
 find it.** `PhoneScrcpy.feedback(message, ok)`, `PhoneCamera.errorOccurred(message)` and
 `PhoneMic.errorOccurred(message)` have been raised on every failure since those services landed,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,18 @@ own repo; the installer pins which revision it builds.
   picker with the real-scene thumbnail, and the Quality dial all appear once
   it is installed (Settings > Update Dots). On the previous renderer those
   controls stay hidden.
+- **Lyrics load faster and stop re-fetching.** Fetched lyrics are cached on
+  disk, so replaying a song or reopening the media sidebar no longer re-hits
+  the network; a track change now fetches once instead of two or three times,
+  and a hung fetch gives up after 20s instead of spinning forever.
+
+### Fixed
+- **Lyrics no longer flicker "no lyrics" on every track change**, and seeking
+  within a song no longer disrupts them.
+- **Media cover-art tint no longer freezes grey** on the dock after a track
+  with no artwork, and progress bars no longer misbehave on live streams or
+  when no player is present. Cover art for local files and on slow IPv6
+  networks loads reliably in the sidebar.
 
 ## [1.0.0-rc-13] — 2026-09-04
 
--- a/dots/.config/quickshell/imi/modules/common/widgets/WaveVisualizer.qml
+++ b/dots/.config/quickshell/imi/modules/common/widgets/WaveVisualizer.qml
@@ -29,9 +29,13 @@ Canvas { // Visualizer
         var n = points.length;
         if (n < 2) return;
 
-        // Smoothing: simple moving average (optional)
+        // Smoothing: simple moving average (optional). Reuse the buffer across
+        // paints - reallocating it every frame (this runs at ~display rate per
+        // visible visualizer) churned the GC for nothing; only the band count
+        // changing needs a new array.
         var smoothWindow = root.smoothing; // adjust for more/less smoothing
-        root.smoothPoints = [];
+        var sp = root.smoothPoints;
+        if (sp.length !== n) { sp = new Array(n); root.smoothPoints = sp; }
         for (var i = 0; i < n; ++i) {
             var sum = 0, count = 0;
             for (var j = -smoothWindow; j <= smoothWindow; ++j) {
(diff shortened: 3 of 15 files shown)


XephyLon:
Read all four commits against `main` (80f1fe897). Checks green, mergeable. Two defects worth fixing before merge, the rest nits.

### Defects

1. **A transient network failure is cached as "no lyrics" for six hours** (`lyrics.py`, `cache_put(key, "not_found")` at the end of `main`). `http_json` returns `None` for every exception, so an offline moment, a DNS hiccup or a Cloudflare 403 walks all five providers to `None` and lands in the negative cache exactly like a real miss. The shell's 20 s watchdog does not help here: fast failures (no route, DNS refused) finish in milliseconds and reach `cache_put`. Result: the song that was playing when the laptop lost Wi-Fi shows no lyrics for the rest of the evening, and the reopen-is-instant behaviour makes it look deliberate. Fix: distinguish "a provider answered and had nothing" from "no provider could be reached" (a module-level flag set in `http_json`'s except, or return a sentinel), and only negative-cache when at least one provider answered. A test that stubs `http_json` to raise and asserts nothing was written closes it.
2. **The dock's blurred art re-decodes on every frame of the card's width animation** (`DockMedia.qml`, `sourceSize.width: Math.round(card.width)`). `card` is anchored to `root`, and `root` has `Behavior on implicitWidth` for the show/hide, so `sourceSize` changes ~12 times per animation and each change is a full reload. This is the bound-to-geometry reload trap AGENT.md names in the wallpaper decode entry. Bind the bound to the constants instead: `root.cardWidth` and the card's resting height (or a fixed 240x46), which is what the comment on the change intends anyway.

### Nits

- `WaveVisualizer.qml`: on the resize path, `root.smoothPoints = sp` copies the (still empty) JS array into the `list<var>` property, and the `sp[i] =` writes that follow land in the detached array, so the paint after every band-count change reads `undefined` and draws nothing for one frame. Fill first, assign after; the steady-state path is fine because the wrapper writes through, which is what the old `push` relied on too.
- `MediaArtSource.qml`: `onExited` sets `downloaded = true` regardless of curl's exit code, and without `-f` a 404 body is saved as the "art". Pre-existing in all four copies, but now there is one place to fix it: `curl -4 -fsSL` and `downloaded = exitCode === 0`.
- `LyricsService.qml`: `_ignoreNextExit` is set before every kill guarded by `lyricsProc.running`. If the old process has already exited but its `exited` has not been delivered, the flag correctly absorbs it; if it had fully exited, no kill happens and no flag is set. Looks right, just noting the watchdog is what bounds the residual case (a swallowed exit on a fetch that dies silently) at 20 s.
- Docs receipt says no AGENT.md mechanism changed. `MediaArtSource` is a new one-element-per-purpose rule (four drifted copies collapsed into one) of the kind AGENT.md usually records, so a future media surface does not paste a fifth. Optional.

### Checked and sound

Cache: key normalises case and integer duration, `.json.tmp` + `os.replace` is atomic, prune by mtime bounded to 512, `IMI_LYRICS_NO_CACHE` bypass used by the wire test, TTLs tested both ways. Multi-stamp LRC fix matches the format (inline stamps are absolute), tested. Debounce + `lastKey` dedup: same key in `loading` or `ok` keeps the run; hidden view clears state and key. Watchdog stops on stdout and on exit. `Lyrics.qml` claims `overridePlayer` only while visible and releases on destruction. Every progress binding now guards `length > 0` (NaN with no player, Infinity on a live stream). Removed properties (`Media.borderless`, `DockMedia.buttonPadding`, the sidebar's visualizer props) have no remaining readers; `Dock.qml:315` still passes `buttonPadding` to `DragApps`, which keeps its own. Left-sidebar position poll gated on `sidebarLeftOpen`. Changelog entries present.

XephyLon:
All addressed (`fcfa6912`, `8701209e`, `28840af3`).

### Defects
1. **Network failure poisoned the negative cache** — a miss is now negative-cached only when at least one provider actually ANSWERED (`net_answers`, bumped on every parsed response — error-shape included — in `http_json` and the cubey fetch). Zero answers prints `not_found` and writes nothing, so the next fetch retries. Tested both ways: an unreachable stub writes nothing; an answering-but-empty stub caches.
2. **DockMedia blur re-decoded per animation frame** — the decode bound now reads the card's resting constants (`root.cardWidth` × 46), not the animated `card.width/height`.

### Nits
- **WaveVisualizer resize blank frame** — fill first, assign after; steady-state keeps writing through the wrapper.
- **MediaArtSource 404-as-art** — `curl -4 -fsSL` and `downloaded = exitCode === 0`.
- **`_ignoreNextExit` residual** — agreed as analyzed; the watchdog bounds the swallowed-exit case, left as is.
- **Docs** — took the optional: AGENT.md's "two names for one thing" section now records MediaArtSource as the one cover-art pipeline (with the why-not-`common/widgets` note), and the PR's Docs receipt is updated to match.

Full suite green after the changes.

XephyLon:
Round 2 (fcfa6912a, 8701209e9, 28840af30) read. Both defects and all four nits are addressed as described; checks green; the AGENT.md citation resolves and `media_art.js` exists where the entry points.

One refinement on the negative-cache guard, not blocking:

- `http_json` counts an answer only after a successful parse, but `urlopen` raises `HTTPError` for any 4xx/5xx, and LRCLIB's real "no match" is a 404. So a genuine miss where LRCLIB 404s and the JSON providers also miss still counts, but one where only LRCLIB is reachable and answers 404 does not, and that song is re-fetched every play. The safe direction, just wasteful. Catch `urllib.error.HTTPError` separately and bump `net_answers` there too: a status line is the server answering. `URLError`/timeout stays "no answer".

Otherwise ready to merge from my side.

PR #359, opened 2026-09-07
fix(wallpaper): the crop picker keeps its drag from the sidebar's Flickable

XephyLon:
## Summary

Dragging the Fill crop box in the Wallpaper Engine sidebar scrolled the whole settings column along with the box (recording 2026-09-07 07:51; frames 2 s apart show the column shifted up while the box moved). The picker is a `MouseArea` inside the sidebar's `StyledFlickable`, and a Flickable steals any press once the pointer travels past the drag threshold. `preventStealing: true` keeps the gesture with the picker from the press. `StyledFlickable`'s own MouseAreas take no buttons (wheel only), so nothing else contends for it.

`test_we_overrides_wiring.py` pins the property on that MouseArea and plants its removal in its can-fail self-test.

## Test plan

- `test_we_overrides_wiring.py`: 2/2 here; against `main`'s tree it fails with "the crop picker's MouseArea does not keep its drag from the Flickable".
- Full suite: first run 2026-09-07 failed only in the WidgetResizeGrip nested-weston harness ("A connection to the bus can't be made" from its private session bus, unrelated to this change); that test passes alone (33 s) and a full rerun is in progress, result to follow in a comment.
- Live check after deploy (needs a real pointer, so yours): open the WE sidebar for a project, drag the crop box up and down - the column must stay put.

Docs: not needed — one-property gesture fix inside an existing component; no AGENT.md rule changes and the wiring test's docstring already covers the sidebar contract.
Changelog: updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ own repo; the installer pins which revision it builds.
 
 ## [Unreleased]
 
+### Fixed
+- **Dragging the Wallpaper Engine crop box no longer scrolls the sidebar.**
+  A vertical drag on the Fill crop picker moved the box and scrolled the
+  whole settings column with it; the picker now keeps the drag to itself.
+
 ### Changed
 - **The installer stops hoarding old renderers and caches.** Each Wallpaper
   Engine pin used to leave the previous ~1.4 GB prebuilt behind (7 of them,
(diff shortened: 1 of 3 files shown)


XephyLon:
Full suite rerun on the branch worktree: 1680 passed, 0 failed, 0 skipped (2026-09-07). The first run's only failure was the WidgetResizeGrip nested-weston harness losing its private session bus; that test passed alone and in this rerun.

PR #364, opened 2026-09-09
AI: dictation - a mic in the composer, local transcription

XephyLon:
Implements `docs/proposals/ai-voice-input.md`.

A mic chip in the composer (or `qs ipc call ai dictate toggle` on a key) records the default source until pressed again and puts the transcript in the composer for editing - or sends it at once with the auto-send switch. `scripts/ai/ai_dictate.py` records with `pw-record`/`parec` and transcribes locally with `faster-whisper` (in the shell's venv or the system python) or whisper.cpp's `whisper-cli`; the `provider` engine (OpenAI-compatible transcription, keyed like the chat) is opt-in and labelled as audio leaving the machine. A model is downloaded only from the Settings button, never on first use. `AiDictation` owns idle → listening → transcribing, the clock, the watchdog (`ai.dictation.maxSeconds`) and the draft hand-off; the composer re-reads the draft on `transcribed`. The privacy indicator lights for the whole recording. New singleton: a full `qs` restart is needed once after deploying.

**On this machine:** no transcriber is installed yet; the chip says so. `uv pip install faster-whisper` into `~/.local/state/quickshell/.venv`, then Settings > Services > AI > Dictation > Download model.

**Tests:** `tests/test_ai_dictate.py` (the state machine with a stub recorder and the fake-transcript seam), `AiDictationRuntimeTest.qml` / `tests/test_ai_dictation_runtime.py` end to end in a real shell (12 checks). Full suite: see the last comment.

Docs: updated AGENT.md (the dictation point)
Changelog: updated


XephyLon:

PR #365, opened 2026-09-09
AI: local retrieval over your own documents

XephyLon:
Implements `docs/proposals/ai-local-rag.md`.

Name folders under Settings > Services > AI > Documents and press Index now; the model gains a `search_documents` tool (all four dialects) that answers with the matching passages in a labelled data block and queues the source files as chips under the reply, and a Documents chip in the composer attaches matching passages to every message (they ride in the wire content while the bubble keeps the typed text). Everything stays local: the default embedder is an offline keyword index; `ollama:<model>` uses the daemon's `/api/embed`; vectors are namespaced per embedder.

The privacy contract lives in `scripts/ai/ai_rag.py`, where a test can reach it: only the named folders (real paths), dotfiles never, the home's key and config directories never even when named, `.noindex` subtrees and `.gitignore`d names skipped, binaries and files over 2 MB skipped, incremental by mtime+size, removing a folder forgets its rows at once. Nothing watches user folders. New singleton: a full `qs` restart is needed once after deploying.

**Tests:** `tests/test_ai_rag.py` (the contract against a temp tree and a fake `/api/embed`), `AiRagRuntimeTest.qml` / `tests/test_ai_rag_runtime.py` end to end in a real shell (10 checks). Full suite: see the last comment.

**To try:** add `~/Documents` (or a notes folder), Index now, then ask the assistant something only those files answer. Pull `nomic-embed-text` from the Ollama page (#366) and switch the embedder for semantic matches.

Docs: updated AGENT.md (the local retrieval point)
Changelog: updated


XephyLon:

PR #367, opened 2026-09-09
Launcher: an Ask row for the assistant under @, and an opt-in fallthrough

XephyLon:
Implements `docs/proposals/ai-in-overview.md` (the prefix, the Ask row, the opt-in fallthrough; inline answers deferred).

Type `@` and a question in the overview to get one "Ask <model>" row first; Enter closes the overview, opens the left sidebar on the Intelligence tab and sends the question (`LauncherSearch.askAssistant`). With `search.ai.fallthrough` on (Settings > Services > Search, default off because it changes what Enter does on a miss), a query of four or more words that matched no app, pack, setting or action gets the same row last, ahead of only the default command/math/web rows. The row is built only while the selected model is usable (`Ai.currentModelHasApiKey`), so the launcher never offers a dead end, and nothing is sent before Enter. The prefix is `@` (`?` was already web search) and configurable; the search bar's type/icon and the query highlight follow.

**Tests:** `tests/test_launcher_ai_ask.py` (config keys and defaults, the usable-model gate, no send while typing, ordering, `resultInputs` observation); `LauncherAskRuntimeTest.qml` / `tests/test_launcher_ask_runtime.py` drive a real shell with a keyless probe model (10 checks). Full suite: see the last comment.

**To try:** Super, type `@what is a wayland compositor`, Enter. Then Settings > Services > Search, enable the fallthrough switch, type a long sentence that matches nothing.

Docs: updated AGENT.md (the launcher's Ask row point)
Changelog: updated


XephyLon:
Full suite on this branch: 1680 QML tests passed, 0 failed, all 272 stages green (local run_tests.sh, 2026-09-14).

PR #370, opened 2026-09-09
Overlays: an enter/leave lifecycle for the desktop menu, screenshot toast and drop shelf

XephyLon:
Implements `docs/proposals/overlay-motion-tier.md` for the three overlays that still popped.

`Appearance.animation` gains `overlayEnter` (300 ms) and `overlayExit` (180 ms), both through the motion policy. `OverlayLifecycle` (modules/common/widgets) is the one mechanism: `wanted` follows the open flag, `alive` is what the surface's Loader binds to and outlives the flag by the leave, `progress` is the single scalar the card's opacity and scale read, `closed()` fires when the leave has finished. The desktop menu, the screenshot toast and the drop shelf use it: they no longer vanish on the frame the flag drops, their input mask follows the flag so a leaving overlay never eats a click or a drop, and the drop shelf gains the entrance it never had. The session screen already had the same shape; the cheatsheet is a toplevel the compositor animates.

**Tests:** `tests/lint_overlay_lifecycle.py` fails a host whose surface `active:` reads the flag or that has no flag-gated mask; `OverlayLifecycleRuntimeTest.qml` / `tests/test_overlay_lifecycle_runtime.py` measure the sequence against the real catalogue, reduce-motion included (15 checks). Full suite: see the last comment.

**To try:** right-click the desktop and dismiss the menu; take a screenshot and let the toast time out; summon the drop shelf and close it. Then Settings > Appearance > reduce motion, repeat.

Docs: updated AGENT.md (the overlay lifecycle point)
Changelog: updated


XephyLon:

PR #371, opened 2026-09-09
Headless widgets: the study, and Phase 1 step 1 (NetworkSpeed logic extracted)

XephyLon:
The headless-by-design study (`docs/proposals/headless-widgets.md`, moved from the orphan `claude/headless-widgets-feasibility` branch, content unchanged) plus the first step of its Phase 1.

`modules/imi/bar/NetworkSpeed.qml`'s logic - the `/proc/net/dev` parser, the sample-to-rate state machine, the rate label - moves to `modules/imi/bar/network_speed.js` as pure functions with `tests/tst_network_speed.qml` (parser sums every interface but lo and skips malformed lines; the first sample is a baseline; a counter reset is no traffic, not a negative burst; a repeated timestamp changes nothing; labels). The widget keeps the FileView, the poll and one state object; its rate properties are read-only views of that state. No behaviour change. The remaining rows of the backlog table follow the same pattern, one per PR.

Docs: not needed — a pure refactor with its own unit test; the study doc records the step
Changelog: not user-visible — refactor, no behaviour change


XephyLon:
Full suite on this branch: 1687 QML tests passed (the 7 new tst_network_speed.qml cases included), 0 failed, all stages green (local run_tests.sh, 2026-09-14).

PR #376, opened 2026-09-14
feat(launcher): one-sentence answers under the Ask row

XephyLon:
## What

Approach 3 of `docs/proposals/ai-in-overview.md`: inline answers. With Settings > Services > Search > "Answer short questions inline" on, a `@question` of three or more words gets a one-sentence answer under the Ask row after a pause in typing, streamed from the selected model. Enter carries the answer into the chat, so the conversation continues from it.

- `services/AiInline.qml`: the request, its debounce, the one-sentence system prompt, the 200-character cut, its own dialect strategy instances. Never touches the chat's requester, strategies or messages.
- Gate: `search.ai.inline` (off) AND a usable model AND (loopback endpoint OR `search.ai.inlineWithCloud`, off). Keyless is not local: a remote keyless server still receives keystrokes.
- `LauncherSearch` drives it from `onQueryChanged` the way the qalc spawn is driven, never from the results build; `SearchItem` binds the Ask row's subtitle straight to `AiInline`, so a streaming answer never rebuilds the list. A keystroke, a query without the prefix, or the overview closing cancels.
- Settings: the switch, and (while it is on) the cloud switch and the pause before asking.
- `Ai.addModel` parents the new model on `root`, not `this`: called through the singleton from another QML context it segfaulted in Qt 6.11's overload resolution of `createObject`. `addMessage` already used `root`.
- The JS dialect lint now bans `trimEnd`/`trimStart`/`replaceAll` (QV4 has none; the cut answer threw on it in the nested shell).

## Tests

- `tests/test_ai_inline_contract.py` (7): defaults off, the gate, driven from the query handler and never named by the builder, the row bound to the singleton, Enter carrying the answer, bounded and isolated.
- `tests/test_ai_inline_runtime.py` + `AiInlineRuntimeTest.qml` (15 checks) against a fake OpenAI-compatible streaming server in a nested shell; the python side counts exactly three requests, each with the one-sentence prompt, the question alone, no tools.

Full `run_tests.sh` in the PR worktree: result in the final comment.

## Live check

Needs a local model (Ollama) selected, or the cloud switch. Type `@` + a question of three or more words in the launcher, pause, watch the subtitle; press Enter and the chat opens with the question and the answer already there.

Docs: updated AGENT.md (a point on inline answers before the OllamaCatalog point) and CHANGELOG.md.
Changelog: updated (Unreleased, Added).


XephyLon:
Review gate: 9.2 overall after three rounds (performance 9.5, look & feel 9.0, design 9.0); every deduction taken.

Full `tests/run_tests.sh` in the PR worktree: `EXIT=0` (285 stages).

PR #378, opened 2026-09-14
feat(config): appearance on its own file (config split, stage 1)

XephyLon:
## What

Stage 1 of `docs/proposals/config-storage-split.md` (step 6 of its plan): `Config.options` becomes an aggregator and `appearance`, the largest domain and the busiest writer, moves to its own file, `config.d/appearance.json`. Every other domain still shares `config.json`. No `Config.options.x.y` read or write anywhere changes.

- `Config.qml`: a `QtObject` aggregator with one alias per top-level domain (`Config.domains` lists them). The appearance block sits on a second `JsonAdapter` behind its own `FileView`, with its own reload/write timers and the shared `readWriteDelay`, `blockWrites` and `configDirTimedOut` gates. `ready` waits for both files; the migrations run once both are in.
- The split happens once, on the first load that finds `appearance` in `config.json` and no split file. **Measured while building it: the main adapter's next write drops a key it no longer declares**, so the downgrade path is a copy taken first, `config.json.pre-split-<date>` (`cp -n`, waited for), then the object becomes the file verbatim.
- Outside QML: `switchwall.sh` (every `appearance.*` read and the accent-colour write), `applycolor.sh` (and the terminal background script's `--config`) and `presets.sh` go to the split file when it exists. A preset stays one shared document: `--save` folds the file in, `--apply` splits it back out.
- `Directories` creates `config.d` at startup. This PR folds in the proposal document (#369's commit) and marks its stage 1 landed.

## Tests

- `tests/test_config_split_contract.py` (3): every declared domain aliased and listed, `ready` waits for both files, the split copies before it writes and never creates the file after a timed-out directory migration, the scripts read (and switchwall writes) through the split file.
- `tests/test_config_split_runtime.py` + `ConfigSplitRuntimeTest.qml`, a real shell twice: an old-style `config.json` gets split (value read, file created, an appearance write lands there, a bar write lands in `config.json`, one pre-split copy predating the session's writes); a split file already present beside a stale copy wins, no copy taken.
- `test_presets.py` round-trips a preset through the split files; every test that parses `Config.qml` (44) and every test that reads the three scripts is green; `test_ai_tools_runtime.py` (the `get_shell_config` tool walks the aggregator) and `test_config_dir_migration_runtime.py` green; nested shell load green.

Full `run_tests.sh`: result in the final comment.

## Downgrade

Hand `config.json.pre-split-<date>` back as `config.json` and remove `config.d/`; the CHANGELOG says so.

Docs: updated AGENT.md §The Config system (point 3, the split and how a further domain follows) and docs/proposals/config-storage-split.md (status + the measured migration note); CHANGELOG.md.
Changelog: updated (Unreleased, Changed).

Closes #369.


XephyLon:
## Measurements, corrected (the review asked for arithmetic that closes)

Same harness, same seeded document (shipped defaults + two probe values), one appearance write and one bar write after `ready`; sizes read from disk afterwards:

| | main (control) | this branch |
|---|---|---|
| config.json after the writes | 38,130 B | 31,056 B |
| config.d/appearance.json after the writes | n/a | 7,077 B |
| sum | 38,130 B | 38,133 B (the difference is the one `"appearance":` key) |
| bytes rewritten by one appearance write | 38,130 | 7,077 (5.4x less) |
| bytes rewritten by one bar write | 38,130 | 31,056 |
| `Config.ready` (250 ms tick) | 500 ms | 500 ms in all three modes (normal, splitting, arriving upstream config) |

The earlier table's branch column was an estimate from `defaults/config.json`; these are the files the adapters wrote. The ready rows say only "within the same 250 ms tick", not "identical".

XephyLon:
Full suite (tests/run_tests.sh), second run on d1f817600 after the FileView mock gained `saved`/`saveFailed`: EXIT=0, no failures. Ran with the fake local Ollama on 11434 stopped first.

PR #379, opened 2026-09-14
feat(frame): frame mode, stage 1 (geometry, bands, fillets)

XephyLon:
## What

Stage 1 of `docs/proposals/frame-mode.md` (#67): geometry first. Behind `appearance.frame.enable` (off by default), the shell draws the bar, a band along the other three screen edges and rounded inner corners as one connected surface in the bar's colour.

- `services/FrameGeometry.qml` + `frame_geometry.js`: the single geometry authority. Band thickness = `appearance.frame.thickness`, or the compositor's outer gap when 0 (the band then fills exactly what windows already leave). The bar's edge is the bar; each inner fillet sits where two occupied strips meet.
- `modules/imi/frame/Frame.qml`: the two side bands and the one opposite the bar, per screen. Top layer, `ExclusionMode.Ignore`, empty mask: nothing reserved, no input, hidden with the bar's fullscreen rules.
- `ScreenCorners` moves its four fillets to the inner corners and paints them the frame's colour; `BarContent` squares the plate. Nothing outside the authority computes an inset (pinned).
- Settings > Appearance > Frame: the switch, the thickness (rowVisible). Vertical bar: not framed in this slice (`FrameGeometry.enabled` is false then). Modals docking into the frame and per-screen frames are the next slices; this PR folds the proposal document in and leaves #67 open.

## Verified

- `tst_frame_geometry.qml` (3), `test_frame_mode_contract.py` (4), nested shell load 3/3, settings/search index and row lints green.
- Sandbox (nested Hyprland, gaps_out 5, Hug bar, the shipped wallpaper, two terminals): bar plate square and full width, bands on left/right/bottom in the bar's colour, inner fillets at all four corners, windows sitting inside the frame. Captures in the review thread.

Full `run_tests.sh`: result in the final comment.

## Review follow-ups (not in this slice)

The fifth review pass scored 8.7 (performance 9.0, look & feel 8.5, design 8.5) and left four items for the next slice:

- A monitor excluded by `bar.screenList` still gets a bar-shaped inset (the bar's zone is one global value); `barReservesOn(screenName)` on the authority, reading the same list the bar's `Variants` filters on.
- `FrameGeometry` lives in `services/` and imports `qs.modules.imi.dock` for `DockReservation` - the only services→feature import in the tree; move `FrameGeometry.qml` + `frame_geometry.js` to `modules/imi/frame/` and add a lint that nothing under `services/` imports `qs.modules.imi.*`.
- (done in this PR) the probe re-arms through a `probeArmed` flag instead of writing over its `running` binding; the per-screen dock predicate is memoised once per authority.

Docs: updated AGENT.md (a point on the geometry authority), docs/proposals/frame-mode.md (status), CHANGELOG.md.
Changelog: updated (Unreleased, Added).


XephyLon:
Full suite (tests/run_tests.sh) on 8ae8a820b: EXIT=0, no failures. Two earlier runs each tripped one stale contract that the frame changes had made true in a new place (the dock zone derivation now read through DockReservation; the float-islands zone term now living in Appearance.sizes.barReservedHeight); both tests were updated, nothing in the shell changed between runs. Fake local Ollama stopped before each run.

PR #399, opened 2026-09-17
fix(sandbox): the review sandbox's stop ends the whole session

XephyLon:
The review sandbox's `stop` left sessions running, and over a day of review sandboxes that corrupted the measurements a review depends on. This fixes the tooling on its own, stacked on #398 so the three branches fast-forward in order.

## What was wrong
- `start` recorded the pid of the subshell around `cd && qs &`, so `stop` killed the subshell and the shell sometimes survived.
- `stop` never ended anything the shell had started: the tray watchdog, nmcli and busctl monitors, clight, a keyring, the session's D-Bus and at-spi. 274 of them had piled up, and each tray watchdog whose bus was gone spun at 14% of a core. The machine sat at a load average of 73, and a dock review's idle CPU swung tenfold between starts in a way that looked like a property of the build.
- `start` guessed a directory was a sandbox from file names, so it could have wiped a project whose Python venv is called `env/`.

## What `stop` and `start` do now
- Every process of a session carries `IMI_SANDBOX_SESSION=<sandbox-dir>`. The env file does not export it, so a terminal that sourced the env file is never matched.
- `stop` ends every marked process, the shell first (found by its argv[0], not a self-matching `pgrep -f`).
  - It needs no env file, and never kills a recorded pid on its own.
  - It removes only a `/tmp/imi-sb-XXXXXX` run dir that `start` made, after lazily unmounting any FUSE mount left there.
  - It refuses to run from inside a sandbox, reports what it ended and whether anything is left, and exits non-zero if anything is.
- Both commands canonicalise the sandbox path, and the scan matches it as a fixed string.
- `start` writes an `.imi-sandbox` sentinel and treats only that as a sandbox.
  - It stops a sandbox it is about to reuse, and gives up if it cannot.
  - It refuses any other non-empty directory.
  - A start that fails ends the session it began.
- Two hooks exist for the test, `SANDBOX_START_WAIT` and `SANDBOX_STOP_KILL`. Both are validated and announce themselves when set.

## Tests
`test_sandbox_shell.py` is new and wired into `run_tests.sh`. It drives start and stop against fake `Hyprland`, `dbus-run-session` and `qs` binaries: 17 tests, about 29 s.
- The recorded pid is the shell.
- Stop ends the whole session and reports it.
- A shell deaf to SIGTERM is killed.
- A sourced terminal survives.
- A stranger behind a stale pid survives.
- A session with no env file is ended.
- Foreign or empty `run.path` values are ignored.
- Start over a live sandbox stops it first.
- Stop from inside a sandbox refuses.
- Path spellings and literal matching are handled.
- Both start refusals hold.
- A failed start ends its session and removes its run dir.
- An incomplete stop fails.
- The hooks are validated and announced.
- No test leaks a run dir.

On a real sandbox, `stop` reports "26 processes ended, nothing left" every time.

## Review
This tooling was split out of the dock branch at the reviewer's request. Scores (performance / ergonomics and output / design compliance):

| Round | Score | Fixed after |
|---|---|---|
| 8 | 8.5 / 8.0 / 8.0 | stop needing an env file; unvalidated `run.path`; start wiping a live sandbox; no full-path argv[0] test; no report |
| 9 | 9.0 / 8.5 / 8.5 | path spelling, literal match, start ignoring a failed stop, incomplete stop exiting 0 |
| 10 | 8.5 / 8.5 / 8.0 | a venv-shaped directory counted as a sandbox; the test deleting other sandboxes' run dirs; `"$0"`; silent hook |
| 11 | 9.0 / 8.5 / 8.5 | hook validation, refusal hint, failed start's run dir asserted (done after the round) |

The full suite ran once, on the top of the stack (`feat/split-shader`, which contains this branch); the result is in that PR.

Docs: updated AGENT.md §review sandbox (the sandbox point), tests/sandbox/README.md
Changelog: not user-visible — review tooling under tests/sandbox only


XephyLon:
Landed in main, despite the closed label. The stack (#398, #399, #400) was merged as one fast-forward of main to the top of it, so all 26 commits of this branch are in main (main is now c50341fd5). This PR reads as closed rather than merged only because its base branch `feat/motion-split` was deleted after the fast-forward instead of being retargeted to main first. The sandbox stop fix and its 17-test suite are on main.

Issue #70, opened 2026-08-02
Changing wallpaper doesn't work


apollo79:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://ii.clsty.link/en/ii-qs/04troubleshooting/) and [Usage](https://ii.clsty.link/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://ii.clsty.link/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

<details><summary>Quick diagnose</summary>
/========================\
| Checking git repo info |
\========================/

[===diagnose===] git remote get-url origin
https://github.com/XephyLon/immaterial-impulse.git
[---SUCCESS---]

[===diagnose===] git rev-parse HEAD
8212c45ada3001a72082e4d9652298020e604a61
[---SUCCESS---]

[===diagnose===] git status
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean
[---SUCCESS---]

[===diagnose===] git submodule status --recursive
[---SUCCESS---]

/=================\
| Checking distro |
\=================/

[===diagnose===] _check_distro_id
distro ID: fedora
distro ID_LIKE: 
[---SUCCESS---]

[===diagnose===] cat os-release
cat: os-release: No such file or directory
[---EXIT 1---]

/====================\
| Checking variables |
\====================/

[===diagnose===] declare -p XDG_CACHE_HOME
./diagnose: line 36: declare: XDG_CACHE_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_CONFIG_HOME
./diagnose: line 36: declare: XDG_CONFIG_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_DATA_HOME
./diagnose: line 36: declare: XDG_DATA_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_STATE_HOME
./diagnose: line 36: declare: XDG_STATE_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p IMMATERIAL_IMPULSE_VIRTUAL_ENV
declare -x IMMATERIAL_IMPULSE_VIRTUAL_ENV="/home/.../.local/state/quickshell/.venv"
[---SUCCESS---]

/============================\
| Checking directories/files |
\============================/

[...]

PR #400, opened 2026-09-17
feat(dock): the split's neck is a distance field, and a reversal is proportional

XephyLon:
The dock split's neck is now a distance field, built the way the reference shell builds it, and a reversed switch takes a time proportional to how far it got. This is stacked on the sandbox fix, which is stacked on #398.

## Where it comes from
The shell in the reference clip is Clavis ([StatIndet/quickshell](https://github.com/StatIndet/quickshell), GPL-3.0-or-later per its packaging). I read its source after #398 was built.
- It confirms the frame measurement at exactly 0.5 scale.
- Its neck is two rounded-box distance fields joined by a polynomial smooth-minimum, in one fragment shader.
- A merge started mid-split takes `max(220, 820 * progress)` ms.

`docs/proposals/motion-split.md` §1 records the source's constants and files. `licenses/README.md` records what was adapted, from whom, and the Inigo Quilez credit for the distance formulas.

## What changed
- **The neck** is `shaders/split.frag`: the pill's rounded box and the band's half-plane, blended once. It replaces #398's `Shape` path and its one-pixel overlap workaround, which composited to a hairline. Because a flat pill edge faces a flat band, where the source has a circle, the field needed some things the source's does not:
  - The blend tapers along the band, or the neck would let go all at once instead of pinching.
  - Coverage is one device pixel of the field's own gradient. It is taken by central differences (core GLSL ES 1.00, no `fwidth`, per #70) through the taper and clamped to [0.5, 1.5]. Unclamped, it drew a half-covered stalk; with the radius held fixed, a hard-sided post.
  - The pill's field reaches two pixels into the band, less the lift, and the band's edge sits a ramp inside the band. Without either, a hairline or a tinted row showed at the seam.
  - The band's edge is 0 in the box's frame on the top and left edges. A first cut drew a band-coloured slab there.
  - The box is laid out once per motion from the rest margins, so only uniforms change per frame. The window's pixel ratio is passed in, so the ramp holds at fractional scale.
- **Hand-over:** while the field paints, the pill's Rectangle is at opacity 0. This happens only where a shader can paint. The software scene graph and a failed load keep the Rectangle, and there the pill lifts without a neck, with its corners rounding over the whole lift.
- **Corners:** with a neck, the corners start rounding where the pill's ends leave the band (`cornerSpan`), not at the seam.
- **Reversal:** a direction started part way takes the tier times the distance left, never less than the effects tier. The start is latched when the target changes; bound to the moving scalar, the duration shortened itself every frame.
- **Shader binary:** `split.frag.qsb.bake` records the source hash and the `qsb` that baked it. The contract fails on an edit that was not rebaked, and compares bytes when the same `qsb` is installed.
- **Docs:** AGENT.md frame point, M3 §2 Split (the reversal rule and the corner rule), proposal §1 and §6, CHANGELOG, licences.

## Measurements (nested sandbox, 60 fps)
- **Captures:**
  - Default cycle and the corner at 8x.
  - The neck at 6x.
  - Reversals.
  - Pin while floating: no neck.
  - The left edge, and the top edge with the bar at the bottom.
  - Scale 1.25.
  - The software scene graph.
  - The settings row, which this branch does not touch.
- **Pixel scans:** no hairline at the seam, no gap tint, and one body value throughout. The gap under the pill's end is never open while its corner is square, except at the single threshold frame.
- **Reversals:** out and back in 400-650 ms; a landing reversed about 250 ms in comes back in 216-283 ms. Before the latch fix, 680 ms.
- **Cost of the neck alone:** process CPU over 8 s, 721 frames each, three runs.

  | Neck | Hardware GL | llvmpipe |
  |---|---|---|
  | #398's `Shape` | 49-51 | 132-135 |
  | This field | 40-42 | 117-118 |

- **Full shell:** two sets of four interleaved starts on hardware GL.

  | | This branch | #398 |
  |---|---|---|
  | Idle, ticks per 10 s | 59-68 | 58-65 |
  | Eight motions in 12 s | 138-140 | 142-147 |

  The two sets disagree on which idles lower, so that difference is noise. The dock drew 371-376 frames over the motions in both, none over 20 ms.
- **Not measured:** the fragment shader's GPU time on hardware is below Qt's render-loop resolution, and Qt Quick exposes no GL timer query.

## Review (performance / look & feel / design compliance)

| Round | Score | Fixed after |
|---|---|---|
| 5 | 7.0 / 6.0 / 6.5 | top and left band edge offset (slab); `fwidth` on ES 1.00; pill vanishing without a shader; stale docs |
| 6 | 8.0 / 8.5 / 8.5 | per-frame objects; screen pixel ratio; top edge unverified; measurement quality |
| 7 | 8.5 / 9.0 / 6.5 | suite red via two lints; sandbox tooling out of scope (split into its own PR); spill dead area; licence notice |
| 8 | 8.5 / 8.0 / 8.0 | pinch-frame stalk; software corners; stale ranges; `.qsb` freshness |
| 9 | 9.0 / 8.5 / 8.5 | lift pinch post (clamped gradient); fragile byte test (bake record); docs |
| 10 | 9.0 / 8.5 / 8.5 | corner square over the gap at the ends (`cornerSpan`); citations; cost retaken |
| 11 | 9.0 / 8.5 / 8.5 | `cornerSpan` comment, proposal corners point, M3 wording (done after the round) |

Rounds 5-7 were run while leftover sandbox sessions loaded the machine. The tooling PR below this one fixes that, and every figure above was retaken afterwards.

Full suite: ran once, after the review passed, at c50341fd (this branch, which contains the sandbox branch): every section green, 1771 QML checks, exit 0.

Pre-existing and out of scope, to report separately:
- The dock on the bar's edge sits under the bar in frame mode (on main too).
- `sni_watchdog.py` spins when its D-Bus goes away.
- `lint_self_matching_process_patterns` misses `pgrep -u X -f`.

Docs: updated AGENT.md §frame point, docs/M3_GUIDELINES.md §2 Split, docs/proposals/motion-split.md §1 §6, licenses/README.md
Changelog: updated


XephyLon:
Landed in main. The stack was merged as one fast-forward of main to this branch's head, so main is now c50341fd5 (this PR's head commit) and every commit of #398, #399 and #400 is in it. GitHub leaves this PR open because its base is still `fix/sandbox-stop`, which the same fast-forward already made identical to main, so there is no diff left to retarget. Nothing is outstanding: the shader, the proportional reversal, the licence notice and the docs are all on main.

